Issue #15, opened 2018-11-14
Support Neurolucida multiple soma stack


wizmer:
As in:
```("CellBody"
  (Color DarkMagenta)
  (CellBody)
  (   -3.19    -9.26     0.08     0.17)  ;  1, 1
  (   -3.53    -9.60     0.08     2.35)  ;  1, 2
  (   -1.01   -10.61     0.08     2.35)  ;  1, 3
  (    0.50   -11.11     0.08     2.35)  ;  1, 4
 )  ;  End of contour

("CellBody"
  (Color Orange)
  (CellBody)
  (    5.88     0.84    54.62     2.35)  ;  2, 1
  (    6.05     2.53    54.62     2.35)  ;  2, 2
  (    6.39     4.38    54.62     2.35)  ;  2, 3
  (    5.55     5.05    54.62     2.35)  ;  2, 4
  (    2.86     5.72    54.62     2.35)  ;  2, 5
)  ;  End of contour

("CellBody"
  (Color RGB (255, 4, 255))
  (CellBody)
  (    1.85     0.67   -90.22     2.35)  ;  3, 1
  (    0.84     1.52   -90.22     2.35)  ;  3, 2
  (   -4.54     2.36   -90.22     2.35)  ;  3, 3
  (   -6.55     0.17   -90.29     2.35)  ;  3, 4
) ; End of contour
```

mgeplf:
I'm not sure there is a definite standard for this.  I think `z` is
constant comes from this:
https://github.com/neuronsimulator/nrn/blob/99a6a370ef143b3c681eda1deea68bc7e2b8f21d/share/lib/hoc/import3d/read_nlcda3.hoc#L276

However, in the only other multi-`CellBody` reconstructions I could
find, the `z` wasn't constant (unfortunately, I can't share the files
here).

The neuromorpho soma spec mentions [multi-contours here](http://neuromorpho.org/SomaFormat.html):

<img width="1151" alt="image" src="https://user-images.githubusercontent.com/13150119/162684808-38d84b68-8af4-4b5d-8fd2-4269677e43bc.png">

Refering to this file, but it's not in `ASC`, but in the `DAT` format, so I don't know what the `ASC` version looks like.
http://neuromorpho.org/neuron_info.jsp?neuron_name=AKO60sdax2lay

Until we get a definitive example, I'm going to leave this hanging, along w/ the potential solution #306

PR #39, opened 2019-01-04
Support NeuroLucida soma stacks

wizmer:
In NeuroLucida, soma stacks are defined by multiple CellBody S-exps, with
each of them being the soma contour at a given Z altitude.
This commits adds support to this representation by concatenating the points
from each stack into the final vector of points.

Fixes https://github.com/BlueBrain/MorphIO/issues/15


asanin-epfl:
In favor of #306

PR #55, opened 2019-03-13
Vasculature

SnigdhaD:


eleftherioszisis:
Prefix underscores, both single and double, before variable names are considered a bad practice. (https://en.cppreference.com/w/cpp/language/identifiers)
```
An identifier can be used to name objects, references, functions, enumerators, types, class members, namespaces, templates, template specializations, parameter packs, goto labels, and other entities, with the following exceptions:

the identifiers that are keywords cannot be used for other purposes;
the identifiers with a double underscore anywhere are reserved;
the identifiers that begin with an underscore followed by an uppercase letter are reserved;
the identifiers that begin with an underscore are reserved in the global namespace.
"Reserved" here means that the standard library headers #define or declare such identifiers for their internal needs, the compiler may predefine non-standard identifiers of that kind, and that name mangling algorithm may assume that some of these identifiers are not in use. If the programmer uses such identifiers, the behavior is undefined.
```
For class member variables you can use `name_` instead or the hungarian notation one `m_name` to differentiate them from local scope ones. A class function is private then you could list it under the `private` identifier. Python simulates the private behavior with underscores, which `C++` can do with the private section of the class.

wizmer:
> Prefix underscores, both single and double, before variable names are considered a bad practice. (https://en.cppreference.com/w/cpp/language/identifiers)
> 
> ```
> An identifier can be used to name objects, references, functions, enumerators, types, class members, namespaces, templates, template specializations, parameter packs, goto labels, and other entities, with the following exceptions:
> 
> the identifiers that are keywords cannot be used for other purposes;
> the identifiers with a double underscore anywhere are reserved;
> the identifiers that begin with an underscore followed by an uppercase letter are reserved;
> the identifiers that begin with an underscore are reserved in the global namespace.
> "Reserved" here means that the standard library headers #define or declare such identifiers for their internal needs, the compiler may predefine non-standard identifiers of that kind, and that name mangling algorithm may assume that some of these identifiers are not in use. If the programmer uses such identifiers, the behavior is undefined.
> ```
> 
> For class member variables you can use `name_` instead or the hungarian notation one `m_name` to differentiate them from local scope ones. A class function is private then you could list it under the `private` identifier. Python simulates the private behavior with underscores, which `C++` can do with the private section of the class.

This is a totally comment. However @SnigdhaD you can keep it like that for the time being. As all MorphIO is using this wrong convention, I'll have to do a global pass at some point.

wizmer:
And this will need a lot of tests !
Checking what happens with empty vasculatures, cyclic ones, broken ones, not-closed ones...

include/morphio/vascProperties.h
```diff
@@ -0,0 +1,106 @@
...
+{
+    using Type = int;
+};
+
+struct Point
+{
+    using Type = morphio::Point;
+};
+
+struct SectionType
```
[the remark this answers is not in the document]
SnigdhaD: you're right, I forgot about this. I have to add the vasculature enumerations in the enums.h

include/morphio/vascProperties.h
```diff
@@ -0,0 +1,106 @@
...
+    std::vector<float> leakiness;
+};
+
+struct VascSectionLevel
+{
+    std::vector<VascSection::Type> _sections;
+    std::vector<SectionType::Type> _sectionTypes;
+    std::map<uint32_t, std::vector<uint32_t>> _predecessors;
+    std::map<uint32_t, std::vector<uint32_t>> _successors;
+    std::map<uint32_t, std::vector<uint32_t>> _neighbors;
```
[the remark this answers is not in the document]
SnigdhaD: The idea is for ease of use right ? Some one who doesn't want to be bothered about directionality would use neighbors. Actually, I was not sure about if successors and neighbors even need to be exposed right now, since we decided not to make the graph directed for now. In that case, I would make them private and only let the user use neighbors.
@eleftherioszisis do you have any comments ?

src/plugin/vasc_morphologyHDF5.cpp
```diff
@@ -0,0 +1,124 @@
...
+    for (unsigned int i = 0; i < vec.size(); ++i) {
+        const auto& p = vec[i];
+        points.push_back({p[0], p[1], p[2]});
+        diameters.push_back(p[3]);
+    }
+}
+
+void VasculatureMorphologyHDF5::_readSections()
+{
+    auto& sections = _properties.get<VasculatureProperty::VascSection>();
```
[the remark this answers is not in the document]
SnigdhaD: I think this would be different because for neutron/astrocyte the loader reads the starting offset and the parent, here it only reads the offset.

src/vascMorphology.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    std::string extension;
+
+    for (auto& c : source.substr(pos))
+        extension += std::tolower(c);
+
+    auto loader = [&source, &options, &extension]() {
+        if (extension == ".h5")
+            return plugin::h5::load_vasc(source);
+        //if (extension == ".swc")
```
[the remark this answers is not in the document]
SnigdhaD: I'm not sure if @eleftherioszisis has or will have any swc files. There are vtk files and I wonder if we need support for that

SnigdhaD:
Okay, I will resolve the open issues and then work on the tests
@eleftherioszisis Can you mention the corner cases you were testing for VasculatureRepair ?

include/morphio/vascProperties.h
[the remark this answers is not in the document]
wizmer: I haven't said we don't need the public function. I am just talking about the data structure itself.

src/vascMorphology.cpp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+VasculatureMorphology::VasculatureMorphology(VasculatureMorphology&&) = default;
+VasculatureMorphology& VasculatureMorphology::operator=(VasculatureMorphology&&) = default;
+
+VasculatureMorphology::~VasculatureMorphology() {}
+
+bool VasculatureMorphology::operator==(const VasculatureMorphology& other) const
+{
+    if (this->_properties == other._properties)
```
eleftherioszisis: It already returns a boolean, so why not `return this->_properties == other._properties`

src/vascMorphology.cpp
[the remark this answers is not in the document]
eleftherioszisis: Well, there is not an swc spec for the time being.
I would advise against creating a dependency on vtk. There are scripts I made that convert vtk files to h5 which is consumed by MorphIO. In the extraordinary case that more than one labs adopt the vtk format for vasculature datasets, we revisit the issue. :P

include/morphio/vascProperties.h
```diff
@@ -0,0 +1,106 @@
...
+#include <morphio/types.h>
+
+namespace morphio
+{
+namespace VasculatureProperty
+{
+template <typename T>
+void _appendVector(std::vector<T>& to, const std::vector<T>& from, int offset);
+
+struct VascSection
```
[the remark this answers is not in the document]
eleftherioszisis: Are these structs necessary @wizmer. It's a little peculiar to me that we need to redefine every type with a new struct.

eleftherioszisis:
You shouldn't use `this` where it is evident that you use a class member variable. Defining the latter with a postfix underscore `variable_name_` for example (which @wizmer will fix in a following pull), you already have told the reader that this is a class variable. Therefore, `this->variable_name_` is redundant and just makes the reading of the code more difficult in these cases.

https://stackoverflow.com/questions/10310242/in-c-is-it-a-good-practice-to-use-this

include/morphio/types.h
```diff
@@ -23,6 +23,8 @@
 
 // TODO: bcoste fix me
 #include <iostream>
+//#include "iterators.h"
```
wizmer: Remove commented line

include/morphio/vasc/morphology.h
```diff
@@ -0,0 +1,47 @@
...
+private:
+
+    std::shared_ptr<property::Properties> _properties;
+
+    template <typename Property>
+    const std::vector<typename Property::Type>& get() const;
+
+};
+} // namespace vasculature
+} // namespace morphio
```
wizmer: Add a newline

include/morphio/enums.h
```diff
@@ -136,6 +136,17 @@ enum SectionType
     SECTION_ALL = 32
 };
 
+enum VascularSectionType
```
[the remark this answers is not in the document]
wizmer: Good question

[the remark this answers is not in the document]
wizmer: I think so

include/morphio/types.h
```diff
@@ -45,6 +46,11 @@ namespace Property {
 struct Properties;
 }
 
+namespace vasculature {
+    class VasculatureSection;
+    class VasculatureMorphology;
```
wizmer: This class should be Vasculature
I don't remember what did we say about VasculatureSection, should it be Section (is it possible) ?

include/morphio/vasc/morphology.h
```diff
@@ -0,0 +1,47 @@
...
+#include <morphio/vasc/properties.h>
+#include <morphio/types.h>
+
+namespace morphio {
+namespace vasculature {
+class VasculatureMorphology
+{
+public:
+    virtual ~VasculatureMorphology();
+
```
wizmer: Docstring for each method

include/morphio/vasc/properties.h
```diff
@@ -0,0 +1,108 @@
...
+};
+
+struct Point
+{
+    using Type = morphio::Point;
+};
+
+struct SectionType
+{
+    // stores section types for vasculature
```
wizmer: useless comment

include/morphio/vasc/properties.h
```diff
@@ -0,0 +1,108 @@
...
+
+namespace morphio
+{
+namespace vasculature
+{
+namespace property
+{
+struct VascSection
+{
+    // offset
```
wizmer: You can expand on what is this offset

include/morphio/vasc/properties.h
```diff
@@ -0,0 +1,108 @@
...
+};
+
+struct Diameter
+{
+    using Type = float;
+};
+
+struct Connection
+{
+    using Type = std::array<int, 2>;
```
wizmer: Explain this one

include/morphio/vasc/properties.h
```diff
@@ -0,0 +1,108 @@
...
+    std::vector<float> leakiness;
+};
+
+struct VascSectionLevel
+{
+    std::vector<VascSection::Type> _sections;
+    std::vector<SectionType::Type> _sectionTypes;
+    std::map<uint32_t, std::vector<uint32_t>> _predecessors;
+    std::map<uint32_t, std::vector<uint32_t>> _successors;
+    std::map<uint32_t, std::vector<uint32_t>> _neighbors;
```
wizmer: You don't need this data structure.

include/morphio/vasc/properties.h
```diff
@@ -0,0 +1,108 @@
...
+std::vector<Point::Type>& Properties::get<Point>();
+template <>
+std::vector<VascSection::Type>& Properties::get<VascSection>();
+template <>
+const std::vector<VascSection::Type>& Properties::get<VascSection>() const;
+template <>
+std::vector<Connection::Type>& Properties::get<Connection>();
+}
+}
+}
```
wizmer: Newline

include/morphio/vasc/section.h
```diff
@@ -0,0 +1,63 @@
...
+#include <morphio/vasc/properties.h>
+#include <morphio/types.h>
+
+namespace morphio
+{
+namespace vasculature
+{
+class VasculatureSection
+{
+    using SectionId = property::VascSection;
```
wizmer: I don't think you need these `using`

include/morphio/vasc/section.h
```diff
@@ -0,0 +1,63 @@
...
+    using PointAttribute = property::Point;
+
+public:
```
wizmer: Docstrings for all public members

include/morphio/enums.h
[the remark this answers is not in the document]
wizmer: We tried renaming this as SectionType in the vasc namespace, right ? What was the outcome ?

src/iterators.cpp
```diff
@@ -1,8 +1,11 @@
 #include <morphio/iterators.h>
 #include <morphio/section.h>
+#include <morphio/vasc/section.h>
 
 namespace morphio {
+
 class Section;
+//class VasculatureSection;
```
wizmer: remove commented line

src/plugin/vasc_morphologyHDF5.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "../detail/utilsHDF5.h"
+
+#include <iostream>
+
+namespace morphio {
+namespace plugin {
+namespace h5 {
+using namespace vasculature;
+
+property::Properties VasculatureMorphologyHDF5::load(const URI& uri)
```
wizmer: You could rename the file and the class to VasculatureHDF5

src/plugin/vasc_morphologyHDF5.cpp
```diff
@@ -0,0 +1,118 @@
...
+    auto& con = _properties._connectivity;
+    for (size_t i=0; i < vec.size(); ++i) {
+        const auto& c = vec[i];
+        con.push_back({c[0], c[1]});
+    }
+}
+
+}
+}
+}
```
wizmer: newline

src/plugin/vasc_morphologyHDF5.h
```diff
@@ -0,0 +1,45 @@
...
+
+class VasculatureMorphologyHDF5
+{
+public:
+    virtual ~VasculatureMorphologyHDF5();
+    property::Properties load(const URI& uri);
+
+private:
+    void _readDatasets();
+    //HighFive::DataSet _getStructureDataSet(size_t nSections);
```
wizmer: remove commented line

src/vasc/morphology.cpp
```diff
@@ -0,0 +1,125 @@
...
+        int32_t first = connectivity[i][0];
+        int32_t second = connectivity[i][1];
+        successors[first].push_back(second);
+        predecessors[second].push_back(first);
+    }
+
+}
+
+}
+}
```
wizmer: newline

src/vasc/properties.cpp
```diff
@@ -0,0 +1,266 @@
...
+        os << dumpPoint(prop._points[i]) << ' ' << prop._diameters[i];
+        os << std::endl;
+    }
+    return os;
+}
+
+std::ostream& operator<<(std::ostream& os, const Properties& properties)
+{
+    os << properties._pointLevel << std::endl;
+    //os << properties._sectionLevel << std::endl;
```
wizmer: remove

tests/test_4_immut.py
```diff
@@ -2,10 +2,11 @@
 import numpy as np
```
wizmer: test_vasculature.h5 can be renamed vasculature.h5

tests/test_4_immut.py
```diff
@@ -90,3 +91,40 @@ def test_mitochondria():
...
+
+    assert_array_almost_equal(morpho.section(0).diameters,
+                              np.array([1.96932483, 1.96932483, 1.96932483, 1.96932483, 1.96932483,
+                                     1.96932483, 1.96932483, 1.96932483, 1.96932483, 1.96932483,
+                                     1.96932483, 2.15068388]))
+
+    assert_equal(len(morpho.sections), 3080)
+    assert_equal(len(morpho.points), 55807)
+    assert_equal(len(morpho.diameters), 55807)
+    assert_equal(len(morpho.section_types), 3080)
```
wizmer: Don't only test on the size but also on the content.

tests/test_4_immut.py
```diff
@@ -90,3 +91,40 @@ def test_mitochondria():
...
+    assert_equal(len(morpho.section(0).predecessors), 0)
+    assert_equal(len(morpho.section(0).successors), 2)
+    assert_equal(morpho.section(0).successors[0].id, 1)
+    assert_equal(morpho.section(0).successors[1].id, 2)
+    assert_array_equal([sec.id for sec in morpho.sections], range(3080))
+    assert_equal(len([section.id for section in morpho.iter()]), 3080)
```
wizmer: test the content

tests/test_4_immut.py
```diff
@@ -90,3 +91,40 @@ def test_mitochondria():
                        np.array([0, 1, 1, 2], dtype=np.float32))
 
     assert_equal(len(mito_root[1].children), 0)
+
+
+def test_vasculature():
+    morpho = vasculature.VasculatureMorphology(os.path.join(_path, "h5/vasculature.h5"))
```
wizmer: Test also with an empty morphology, a cyclic one, any weird case you can think of.

tests/test_4_immut.py
```diff
@@ -90,3 +91,40 @@ def test_mitochondria():
...
+    assert_equal(len(morpho.points), 55807)
+    assert_equal(len(morpho.diameters), 55807)
+    assert_equal(len(morpho.section_types), 3080)
+    assert_equal(len(morpho.section(0).predecessors), 0)
+    assert_equal(len(morpho.section(0).successors), 2)
+    assert_equal(morpho.section(0).successors[0].id, 1)
+    assert_equal(morpho.section(0).successors[1].id, 2)
+    assert_array_equal([sec.id for sec in morpho.sections], range(3080))
+    assert_equal(len([section.id for section in morpho.iter()]), 3080)
+    all_sections = set([sec.id for sec in morpho.sections])
```
wizmer: test also the section level iterator

tests/test_4_immut.py
wizmer: Test also the equality operator and the not equal one

wizmer: How can this test pass. VasculatureMorphology is undefined, no ?

tests/test_4_immut.py
```diff
@@ -90,3 +91,50 @@ def test_mitochondria():
...
     assert_equal(len(mito_root[1].children), 0)
+
+
+def test_vasculature():
+    morphology1 = vasculature.Vasculature(os.path.join(_path, "h5/vasculature1.h5"))
+    morphology2 = vasculature.Vasculature(os.path.join(_path, "h5/vasculature2.h5"))
+
+    assert_raises(RawDataError, vasculature.Vasculature, os.path.join(_path, "h5/empty_vasculature.h5"))
+
+    assert (morphology1 == morphology1)
```
wizmer:
```suggestion
    morphology1_bis = vasculature.Vasculature(os.path.join(_path, "h5/vasculature1.h5")) 
    assert_equal(morphology1, morphology1_bis) 
```

tests/test_4_immut.py
```diff
@@ -90,3 +91,50 @@ def test_mitochondria():
                        np.array([0, 1, 1, 2], dtype=np.float32))
 
     assert_equal(len(mito_root[1].children), 0)
+
+
+def test_vasculature():
```
wizmer: Split this test into multiple tests

include/morphio/enums.h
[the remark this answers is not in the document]
SnigdhaD: It was returning a conflict, as far as I remember

include/morphio/vasc/section.h
```diff
@@ -0,0 +1,57 @@
...
+
+    const Section& operator=(const Section& section);
+
+    Section(uint32_t id, std::shared_ptr<property::Properties> morphology);
+
+    bool operator==(const Section& section) const;
+    bool operator!=(const Section& section) const;
+    bool operator<(const Section& other) const;
+
+    //bool isRoot() const;
```
wizmer: Remove this

src/CMakeLists.txt
```diff
@@ -1,17 +1,21 @@
 set(MORPHIO_SOURCES
-  morphology.cpp
+        morphology.cpp
```
wizmer: Fix the alignement

include/morphio/iterators.h
```diff
@@ -27,4 +30,25 @@ class Iterator
...
 };
-} // namespace morphio
+
+class graph_iterator
+{
+    friend class vasculature::Section;
+    friend class vasculature::Vasculature;
+    std::set<vasculature::Section> visited;
+
+    std::stack<vasculature::Section> container;
```
wizmer: I think I would prefer a queue rather than a stack to get a breadth first pattern

src/vasc/properties.cpp
```diff
@@ -0,0 +1,291 @@
...
+    const std::string& name, bool verbose_)
+{
+    if (vec1.size() != vec2.size()) {
+        if (verbose_)
+            LBERROR(Warning::UNDEFINED,
+                "Error comparing " + name + ", size differs: " + std::to_string(vec1.size()) + " vs " + std::to_string(vec2.size()));
+        return false;
+    }
+
+    const float epsilon = 1e-6;
```
wizmer:
```suggestion
    const float epsilon = 1e-6f;
```

src/vasc/vasculature.cpp
```diff
@@ -0,0 +1,121 @@
...
+    if (access(source.c_str(), F_OK) == -1)
+        LBTHROW(RawDataError("File: " + source + " does not exist."));
+
+    std::string extension = source.substr(pos);
+
+    property::Properties loader;
+    if (extension == ".h5")
+        loader = plugin::h5::VasculatureHDF5().load(source);
+    else
+        LBTHROW(UnknownFileType("Unhandled file type"));
```
wizmer:
```suggestion
        LBTHROW(UnknownFileType("File: " + source + " does not end with the .h5 extension"));
```

src/vector_utils.cpp
```diff
@@ -143,12 +143,18 @@ std::ostream& operator<<(std::ostream& os,
...
     return os;
 }
+
 std::ostream& operator<<(std::ostream& os, const morphio::Point& point)
 {
     os << morphio::dumpPoint(point);
     return os;
 }
 
+char my_tolower(char ch)
```
wizmer:
```suggestion
// Like std::tolower but accepts char
char my_tolower(char ch) 
```

PR #59, opened 2019-04-04
feature: CMake export for reusability

matz-e:
Some bits that I needed to use MorphIO within TouchDetector.


wizmer:
Same here. I'm glad someone does...

PR #60, opened 2019-04-09
change: allocate vector memory before filling it.

matz-e:
Speeds up morphology loading a bit in my unscientific testing.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -59,6 +59,14 @@ set_target_properties(morphio_obj
   CXX_EXTENSIONS NO
   )
 
+target_compile_options(morphio_obj
+  PRIVATE
+  -Wall
+  -Wextra
+  -Wnon-virtual-dtor
+  -pedantic-errors
+  )
+
 add_library(morphio_static STATIC $<TARGET_OBJECTS:morphio_obj>)
 add_library(morphio_shared SHARED $<TARGET_OBJECTS:morphio_obj>)
 set_target_properties(morphio_shared PROPERTIES OUTPUT_NAME "morphio"
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -63,8 +63,9 @@ target_compile_options(morphio_obj
   PRIVATE
   -Wall
   -Wextra
+  -Werror
   -Wnon-virtual-dtor
-  -pedantic-errors
+  -pedantic
   )
 
 add_library(morphio_static STATIC $<TARGET_OBJECTS:morphio_obj>)
--- a/src/plugin/morphologyHDF5.cpp
+++ b/src/plugin/morphologyHDF5.cpp
@@ -215,13 +215,14 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset)
         std::vector<std::vector<float>> vec(dims[0]);
         dataset.read(vec);
 
-        if (firstSectionOffset < 0) {
-            firstSectionOffset = vec.size();
+        std::size_t offset = vec.size();
+        if (firstSectionOffset >= 0) {
+            offset = firstSectionOffset;
         }
 
-        somaPoints.reserve(somaPoints.size() + firstSectionOffset);
-        somaDiameters.reserve(somaDiameters.size() + firstSectionOffset);
-        for (std::size_t i = 0; i < firstSectionOffset; ++i) {
+        somaPoints.reserve(somaPoints.size() + offset);
+        somaDiameters.reserve(somaDiameters.size() + offset);
+        for (std::size_t i = 0; i < offset; ++i) {
             const auto& p = vec[i];
             somaPoints.emplace_back(Point{p[0], p[1], p[2]});
             somaDiameters.emplace_back(p[3]);
@@ -229,7 +230,7 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset)
 
         points.reserve(points.size() + vec.size() - firstSectionOffset);
         diameters.reserve(diameters.size() + vec.size() - firstSectionOffset);
-        for (std::size_t i = firstSectionOffset; i < vec.size(); ++i) {
+        for (std::size_t i = offset; i < vec.size(); ++i) {
             const auto& p = vec[i];
             points.emplace_back(Point{p[0], p[1], p[2]});
             diameters.emplace_back(p[3]);
@@ -242,21 +243,22 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset)
     vec.resize(_pointsDims[0]);
     _points->read(vec);
 
-    if (firstSectionOffset < 0) {
-        firstSectionOffset = vec.size();
+    std::size_t offset = vec.size();
+    if (firstSectionOffset >= 0) {
+        offset = firstSectionOffset;
     }
 
-    somaPoints.reserve(somaPoints.size() + firstSectionOffset);
-    somaDiameters.reserve(somaDiameters.size() + firstSectionOffset);
-    for (std::size_t i = 0; i < firstSectionOffset; ++i) {
+    somaPoints.reserve(somaPoints.size() + offset);
+    somaDiameters.reserve(somaDiameters.size() + offset);
+    for (std::size_t i = 0; i < offset; ++i) {
         const auto& p = vec[i];
         somaPoints.emplace_back(Point{p[0], p[1], p[2]});
         somaDiameters.emplace_back(p[3]);
     }
 
-    points.reserve(points.size() + vec.size() - firstSectionOffset);
-    diameters.reserve(diameters.size() + vec.size() - firstSectionOffset);
-    for (std::size_t i = firstSectionOffset; i < vec.size(); ++i) {
+    points.reserve(points.size() + vec.size() - offset);
+    diameters.reserve(diameters.size() + vec.size() - offset);
+    for (std::size_t i = offset; i < vec.size(); ++i) {
         const auto& p = vec[i];
         points.emplace_back(Point{p[0], p[1], p[2]});
         diameters.emplace_back(p[3]);
@@ -304,7 +306,7 @@ int MorphologyHDF5::_readSections()
         int firstSectionOffset = vec[1][0];
         sections.reserve(sections.size() + vec.size() - 1);
         bool skipFirst = true;
-        for (auto p : vec) {
+        for (const auto& p : vec) {
             if (skipFirst) {
                 skipFirst = false;
                 continue;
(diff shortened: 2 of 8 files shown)


src/plugin/morphologyHDF5.cpp
```diff
@@ -213,35 +213,53 @@ void MorphologyHDF5::_readPoints(unsigned int firstSectionOffset)
...
         const auto& p = vec[i];
-        if (noNeurites || i < firstSectionOffset) {
-            somaPoints.push_back({p[0], p[1], p[2]});
-            somaDiameters.push_back(p[3]);
-        } else {
-            points.push_back({p[0], p[1], p[2]});
-            diameters.push_back(p[3]);
-        }
+        somaPoints.emplace_back(Point{p[0], p[1], p[2]});
+        somaDiameters.emplace_back(p[3]);
```
arsenius7: At first glance there is certain similarity with the lines above.
Is it possible to address this code duplication?

src/plugin/morphologyHDF5.cpp
```diff
@@ -213,35 +213,53 @@ void MorphologyHDF5::_readPoints(unsigned int firstSectionOffset)
                 "Reading morphology file '" + _file->getName() + "': bad number of dimensions in 'points' dataspace"));
         }
         std::vector<std::vector<float>> vec(dims[0]);
-        // points.resize(dims[0]);
         dataset.read(vec);
-        bool noNeurites = (firstSectionOffset == -1);
-        for (unsigned int i = 0; i < vec.size(); ++i) {
+
+        if (firstSectionOffset < 0) {
+            firstSectionOffset = vec.size();
```
arsenius7: I don't know what's going on here, but maybe it's better to introduce a new variable if we change data type to unsigned here (perhaps it implies the change of the semantics as well?).

src/plugin/morphologyHDF5.cpp
```diff
@@ -283,14 +301,15 @@ int MorphologyHDF5::_readSections()
...
         int firstSectionOffset = vec[1][0];
+        sections.reserve(sections.size() + vec.size() - 1);
+        bool skipFirst = true;
         for (auto p : vec) {
             if (skipFirst) {
                 skipFirst = false;
                 continue;
             }
-            sections.push_back({p[0] - firstSectionOffset, p[1] - 1});
+            sections.emplace_back(Property::Section::Type{p[0] - firstSectionOffset, p[1] - 1});
```
arsenius7: isn't it "almost" same given the move semantics?

src/plugin/morphologyHDF5.h
```diff
@@ -25,7 +25,7 @@ class MorphologyHDF5
     bool _readV11Metadata();
     bool _readV2Metadata();
     HighFive::DataSet _getStructureDataSet(size_t nSections);
-    void _readPoints(unsigned int);
```
arsenius7: Should we add a unit test illustrating the difference between `unsigned int` and `int` in the signature?

src/plugin/morphologyHDF5.cpp
```diff
@@ -213,35 +213,53 @@ void MorphologyHDF5::_readPoints(unsigned int firstSectionOffset)
                 "Reading morphology file '" + _file->getName() + "': bad number of dimensions in 'points' dataspace"));
         }
         std::vector<std::vector<float>> vec(dims[0]);
-        // points.resize(dims[0]);
         dataset.read(vec);
-        bool noNeurites = (firstSectionOffset == -1);
```
matz-e: @arsenius7 the original version tested an *`unsigned int`* for a negative value here. Hence me setting it to `int`.

src/plugin/morphologyHDF5.cpp
matz-e: I don't think we should worry about that just yet. Besides, do `std::size_t` and `unsigned int` differ?

src/plugin/morphologyHDF5.cpp
```diff
@@ -213,35 +213,53 @@ void MorphologyHDF5::_readPoints(unsigned int firstSectionOffset)
...
+
+        if (firstSectionOffset < 0) {
+            firstSectionOffset = vec.size();
+        }
+
+        somaPoints.reserve(somaPoints.size() + firstSectionOffset);
+        somaDiameters.reserve(somaDiameters.size() + firstSectionOffset);
+        for (std::size_t i = 0; i < firstSectionOffset; ++i) {
             const auto& p = vec[i];
-            if (noNeurites || i < firstSectionOffset) {
```
matz-e: Here's an unclean comparison that I removed, too.

src/plugin/morphologyHDF5.cpp
matz-e: I don't think it's worth it really, without some additional work to make it logically seamless.

src/plugin/morphologyHDF5.cpp
matz-e: Wouldn't it require some additional `std::move`? I like to be consistent and just use `emplace_back` for non-primitive types.

src/plugin/morphologyHDF5.h
matz-e: I would like to skip that bit. I just fixed the interface to reflect the internal logic.

arsenius7: I mean, there should have been a unit test exercising  `firstSectionOffset == -1` branch, that should have caught type mismatch.

src/plugin/morphologyHDF5.cpp
arsenius7: `{}` passed to `push_back` was already an rvalue as far as I can tell.
Whatever though.

src/plugin/morphologyHDF5.cpp
arsenius7: OK.

src/plugin/morphologyHDF5.cpp
arsenius7: `firstSectionOffset` is an `int` but is used as `unsigned int` from this point (also compared with `unsigned int` in the subsequent `for`).
It might be that it's worth reflecting this "type change" with an intermediate variable.
I haven't looked into `firstSectionOffset < 0` semantics though.

src/plugin/morphologyHDF5.cpp
arsenius7: That I understood (see also my comment about unit test below).
Btw, wasn't there a compilation warning indicating this comparison?

matz-e: No, that would require some additional compiler flags. I added  those locally just now, need to find a more minimal set than my standard one…

matz-e: @arsenius7 there you go :upside_down_face: 

The old `unsigned int` does not compile now with the new CMake options.

src/plugin/morphologyHDF5.cpp
```diff
@@ -283,14 +301,15 @@ int MorphologyHDF5::_readSections()
         std::vector<std::vector<int>> vec;
         vec.resize(dims[0] - 1);
         dataset.read(vec);
-        bool skipFirst = true;
         int firstSectionOffset = vec[1][0];
+        sections.reserve(sections.size() + vec.size() - 1);
+        bool skipFirst = true;
         for (auto p : vec) {
```
arsenius7: I wonder if `p` is copied from each `vec[i]` and if it has any effect on the performance.

src/plugin/morphologyHDF5.cpp
```diff
@@ -213,35 +213,53 @@ void MorphologyHDF5::_readPoints(unsigned int firstSectionOffset)
...
+        for (std::size_t i = 0; i < firstSectionOffset; ++i) {
             const auto& p = vec[i];
-            if (noNeurites || i < firstSectionOffset) {
-                somaPoints.push_back({p[0], p[1], p[2]});
-                somaDiameters.push_back(p[3]);
-            } else {
-                points.push_back({p[0], p[1], p[2]});
-                diameters.push_back(p[3]);
-            }
+            somaPoints.emplace_back(Point{p[0], p[1], p[2]});
```
arsenius7: Is `somaPoints` a vector of `Point`s here?
Would `emplace_back(p[0], p[1], p[2])` be ambiguous for the compiler then?

src/plugin/morphologyHDF5.cpp
arsenius7: For some reason I still feel leery about re-setting `firstSectionOffset` to the value which gives the same effect as the previous code.
I would suggest to limit this review to unquestionable things like fixing dtype, adding compilation warnings and `.reserve()` calls and see with Benoit about this refactoring.
Alternatively, we can merge it as it is and undo part of it afterwards.

src/plugin/morphologyHDF5.cpp
matz-e: Yes, it is due to the nature of `std::array`.

src/plugin/morphologyHDF5.cpp
matz-e: I'll change that. This could also receive a `const`.

arsenius7:
Most of the changes LGTM; I'll leave it to @wizmer to decide if he would keep the refactoring part, or prefer to revise it.
I guess we'll also need to bump the version.
Thanks for looking into that, @matz-e .

src/plugin/morphologyHDF5.cpp
matz-e: I'd rather have it like this and move the `if` clause outside the loop. I'll merge this now, and we can revert as needed.

matz-e:
A version bump would be appreciated! I see that there are some beta-versions floating around, hopefully we can work something out next week?

src/plugin/morphologyHDF5.cpp
mgeplf: > I don't think we should worry about that just yet. Besides, do `std::size_t` and `unsigned int` differ?

Yes, at least w/ clang 6: https://godbolt.org/z/DU8wad

src/CMakeLists.txt
```diff
@@ -59,6 +59,15 @@ set_target_properties(morphio_obj
   CXX_EXTENSIONS NO
   )
 
+target_compile_options(morphio_obj
```
wizmer: There is already a way to activate Werror (feel free to add additionnal flags there):
https://github.com/BlueBrain/MorphIO/blob/master/.ContinousIntegration.cmake

But I don't think it should be a default for people using the soft (for example through pip install) as opposed to developpers.

matz-e: I'm inclined to disagree here, since this can catch errors when locally making changes (before the CI is even involved). I'd think that in general, you would want to have pip install a pre-built binary anyways?

matz-e:
@wizmer I'm wondering if we could tag a new release with the latest changes, unless you absolutely want to have something reverted?

PR #61, opened 2019-04-17
Remove -Werror to make pip happy.

matz-e:


mgeplf:
What was pip complaining about?

matz-e:
@wizmer pointed out that one some systems/compilers `-Werror` may lead to a failing `pip install`, which will be hard for users to fix/debug.

mgeplf:
Ah, interesting.  Which compilers don't support -Werror @wizmer?

matz-e:
> Ah, interesting. Which compilers don't support -Werror @wizmer?

I don't think that certain compilers don't support `-Werror`, but rather that different compilers apply different standards - a build that may pass for us may fail for a user.

mgeplf:
Yeah, thanks.  -Werror should only be applied in CI, and cause failures there, allowing people to install, and covering us from making mistakes

PR #62, opened 2019-04-17
Testing compilation with Werror in in CI

wizmer:


tristan0x:
Hello @wizmer 
+1 for *-Werror* !

FYI here are the recommended compiler flags: https://github.com/BlueBrain/hpc-coding-conventions/blob/master/cpp/Tooling.md

bob might be helpful: https://github.com/BlueBrain/hpc-coding-conventions/blob/master/cpp/cmake/bob.cmake#L221

wizmer:
Thanks @tristan0x 
I'll add the recommended flags in another PR as I have just tried and this makes a looooot of things to fix.

PR #63, opened 2019-04-25
Make code compliant with more aggressive GCC flags

wizmer:
This flags are only used by Travis CI or if cmake is called with
the option `-DCMAKE_TOOLCHAIN_FILE=../.ContinousIntegration.cmake`


binds/python/bind_misc.cpp
```diff
@@ -109,13 +109,13 @@ void bind_misc(py::module &m) {
...
-        .def_buffer([](morphio::Points &m) -> py::buffer_info {
+        .def_buffer([](morphio::Points &points) -> py::buffer_info {
                 return py::buffer_info(
-                    m.data(),                               /* Pointer to buffer */
+                    points.data(),                               /* Pointer to buffer */
                     sizeof(float),                          /* Size of one scalar */
                     py::format_descriptor<float>::format(), /* Python struct-style format descriptor */
                     2,                                      /* Number of dimensions */
-                    { (int)  m.size(), 3 },                 /* Buffer dimensions */
+                    { static_cast<int>(points.size()), 3 },  /* Buffer dimensions */
```
mgeplf: is this supposed to be ssize_t?

wizmer: It does not compile with `static_cast<ssize_t>

src/vector_utils.cpp
```diff
@@ -77,7 +77,7 @@ std::vector<Point> operator-=(std::vector<Point>& points, const Point& right)
 **/
 float distance(const Point& left, const Point& right)
 {
-    return sqrt((left[0] - right[0]) * (left[0] - right[0]) + (left[1] - right[1]) * (left[1] - right[1]) + (left[2] - right[2]) * (left[2] - right[2]));
+    return sqrtf((left[0] - right[0]) * (left[0] - right[0]) + (left[1] - right[1]) * (left[1] - right[1]) + (left[2] - right[2]) * (left[2] - right[2]));
```
tristan0x:
```suggestion
    return std::sqrt((left[0] - right[0]) * (left[0] - right[0]) + (left[1] - right[1]) * (left[1] - right[1]) + (left[2] - right[2]) * (left[2] - right[2]));
```

binds/python/bindings_utils.cpp
```diff
@@ -6,16 +6,17 @@ namespace py = pybind11;
 
 py::array_t<float> span_array_to_ndarray(const morphio::range<const std::array<float, 3> > &span)
 {
+    const void* ptr = static_cast<const void*>(span.data());
     const auto buffer_info = py::buffer_info(
         // Cast from (const void*) to (void*) for function signature matching
-        (void*)span.data(),                            /* Pointer to buffer */
+        const_cast<void*>(ptr),                            /* Pointer to buffer */
```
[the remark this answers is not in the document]
wizmer: Nope, span is a data structure that holds the coordinate of the immutable morphology. There is no non const accessor.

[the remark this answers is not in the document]
wizmer: But I admit that cast sucks. I wonder if that could lead to a mutation of the data through the python side.

src/plugin/morphologyHDF5.cpp
```diff
@@ -228,14 +238,17 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset)
...
-            points.emplace_back(Point{p[0], p[1], p[2]});
-            diameters.emplace_back(p[3]);
+        if (!noNeurites(firstSectionOffset)) {
+            size_t size = (points.size() + vec.size() -
+                           static_cast<unsigned int>(firstSectionOffset));
+            points.reserve(size);
+            diameters.reserve(size);
+            for (std::size_t i = offset; i < vec.size(); ++i) {
+                const auto& p = vec[i];
+                points.emplace_back(Point{p[0], p[1], p[2]});
```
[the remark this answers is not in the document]
wizmer: Nope, that does not in fact. 
I am getting:
/usr/include/c++/7/ext/new_allocator.h:136:4: error: no matching function for call to ‘std::array<float, 3>::array(const float&)’

CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.2)
 project(MorphIO VERSION 2.0.0)
 
+add_subdirectory(3rdparty/coding-conventions/cpp)
+bob_begin_cxx_flags()
```
wizmer: @tristan0x Is this how it is supposed to be called ?

tristan0x: `bob.cmake` is a bit special, and not ready to be widely used.
It is completely incompatible with ` CMAKE_BUILD_TYPE`. I do not advise you to use it right for now. Sorry!

PR #66, opened 2019-05-02
WIP: cleanup clang8 warnings

mgeplf:


CMakeLists.txt
```diff
@@ -11,6 +11,10 @@ option(${PROJECT_NAME}_CXX_WARNINGS "Compile C++ with warnings" ON)
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   if(${PROJECT_NAME}_CXX_WARNINGS)
     set(FLAGS "${FLAGS} -Werror -Weverything")
+    # Using std=c++11, no need for 98 compat
+    set(FLAGS "${FLAGS} -Wno-c++98-compat")
+    set(FLAGS "${FLAGS} -Wno-c++98-compat-pedantic")
+
     set(FLAGS "${FLAGS} -Wno-documentation")
     set(FLAGS "${FLAGS} -Wno-documentation-unknown-command")
     set(FLAGS "${FLAGS} -Wno-exit-time-destructors")
```
mgeplf: some of these warnings shouldn't be suppressed, IMO

include/morphio/section_base.tpp
```diff
@@ -104,7 +104,7 @@ const std::vector<T> SectionBase<T>::children() const
             result.push_back(T(id_, _properties));
 
         return result;
-    } catch (const std::out_of_range& oor) {
+    } catch (const std::out_of_range&) {
         return result;
```
mgeplf: this strikes me as weird

src/errorMessages.cpp
```diff
@@ -143,7 +143,7 @@ const std::string ErrorMessages::ERROR_MISSING_MITO_PARENT(
 /**
    Return val1 and highlight it with some color if val1 != val2
 **/
-const std::string _col(float val1, float val2)
+static const std::string _col(float val1, float val2)
```
mgeplf: this should actually be moved to a anon namespace

src/properties.cpp
```diff
@@ -4,7 +4,7 @@
 #include <morphio/errorMessages.h>
 #include <morphio/properties.h>
 
-bool verbose = false;
+constexpr bool VERBOSE = false;
```
mgeplf: I think this is from earlier times, is it still needed?

wizmer: It should be removed. But I want to replace it with a helper comparator function that acts the same way as operator== except it display info to the user about what differs.

binds/python/bind_immutable.cpp
```diff
@@ -96,6 +96,7 @@ static void bind_immutable_module(py::module &m) {
                     return py::make_iterator(morpho->depth_begin(), morpho->depth_end());
                 case morphio::IterType::BREADTH_FIRST:
                     return py::make_iterator(morpho->breadth_begin(), morpho->breadth_end());
+                case morphio::IterType::UPSTREAM:
```
wizmer: Why don't you let the default handle it anyway ?

wizmer: Is clang complaining for not all enum values being covered ?

mgeplf: yeah, and it's better to be explicit about what is being skipped - one can't keep all the enum values in their head when they are editing a switch statement, so this way it's obvios what is happening

src/soma.cpp
```diff
@@ -24,6 +24,10 @@ float Soma::volume() const
         return 4 * static_cast<float>(M_PI) * radius * radius;
     }
 
+    case SOMA_SINGLE_POINT:
```
mgeplf: we should be able to figure out a volume for a single point soma, no?

wizmer:
```suggestion
    case SOMA_SINGLE_POINT: {
        float radius = diameters()[0] / 2;
        return 4 / 3 * static_cast<float>(M_PI) * std::pow(radius, 3);
    }
```

PR #67, opened 2019-05-03
Activate clang travis

wizmer:


binds/python/bind_vasculature.cpp
```diff
@@ -12,9 +12,9 @@
 #include <morphio/vasc/properties.h>
 
 namespace py = pybind11;
-using namespace py::literals;
 
-void bind_vasculature(py::module &m) {
+static void bind_vasculature(py::module &m) {
+    using namespace py::literals;
```
mgeplf: why the change?

wizmer: Because clang complains about using `using namespace` in a global scope.

wizmer: And the `static` because it complains about the fact there is no prototype otherwise

mgeplf: Doesn't that `using namespace` trigger this: https://abseil.io/tips/153 ? Or am I reading it wrong?

mgeplf: I guess it only brings in `_a`, so no worries

wizmer: or I can just bring _a with `using py::literals::operator""_a` ?

wizmer: (this syntax is sick by the way, I had no idea you could do that)

PR #76, opened 2019-05-21
Add Morphology::diff method that is like operator=! but verbose

wizmer:


src/properties.cpp
```diff
@@ -209,73 +178,108 @@ bool compare(const T& el1, const T& el2, const std::string& name, bool verbose_)
...
 }
 
 bool CellLevel::operator!=(const CellLevel& other) const
 {
-    return !(this->operator==(other));
+    return diff(other, false);
+}
+
+
+MitochondriaPointLevel::MitochondriaPointLevel(
```
wizmer: This is moved code, not new code

src/mut/morphology.cpp
```diff
@@ -411,38 +411,32 @@ void Morphology::write(const std::string& filename)
         LBTHROW(UnknownFileType(_err.ERROR_WRONG_EXTENSION(filename)));
 }
 
-bool Morphology::_compare(const Morphology& other, bool verbose) const
```
arsenius7: `compare` looked just as good to me (I would also make it a free method, but that's probably a matter of taste).

src/mut/section.cpp
```diff
@@ -203,57 +202,69 @@ std::shared_ptr<Section> Section::appendSection(
...
 }
 
-bool Section::_compare(const Section& other, bool verbose) const
+bool Section::diff(const Section& other, bool verbose) const
 {
     if (this->type() != other.type()) {
         if (verbose)
-            std::cout << "Section type differ" << std::endl;
-        return false;
+            std::cout << "Reason: section type differ" << std::endl;
```
arsenius7: maybe `std::cerr`?

src/mut/section.cpp
```diff
@@ -91,12 +91,11 @@ upstream_iterator Section::upstream_end() const
...
 std::ostream& operator<<(std::ostream& os, Section& section)
 {
-    os << "id: " << section.id() << std::endl;
-    os << dumpPoints(section.points());
+    ::operator<<(os, section);
     return os;
 }
 
-std::ostream& operator<<(std::ostream& os, std::shared_ptr<Section> sectionPtr)
+std::ostream& operator<<(std::ostream& os, const std::shared_ptr<Section> sectionPtr)
```
arsenius7: it's a copy anyway at the moment, so `const` does not change much

src/mut/section.cpp
wizmer: I am not sure. The message is what is expected to be print out, it is not really an error.

src/mut/morphology.cpp
wizmer: good point, I'll change that

src/mut/section.cpp
arsenius7: `stderr` is not necessarily for errors, but for additional "diagnostics" as well.
But now that I took a look at some `diff`-ing tools, all of them use `stdout`, so nvm.

wizmer:
hawk eyes

setup.py
```diff
@@ -44,6 +44,7 @@ def build_extension(self, ext, cmake):
         cmake_args = ['-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=' + extdir,
                       '-DPYTHON_EXECUTABLE=' + sys.executable]
 
+        self.debug = True
```
arsenius7: leftover?

include/morphio/properties.h
```diff
@@ -106,40 +125,30 @@ struct MitochondriaPointLevel
...
+                           SectionRange range);
 
     MitochondriaPointLevel(
         std::vector<uint32_t> sectionId,
         // relative pathlength between the current points
         // and the start of the neuronal section
         std::vector<MitoPathLength::Type> relativePathLengths,
         std::vector<MitoDiameter::Type> diameters);
+
+    bool diff(const MitochondriaPointLevel& other, LogLevel verbose) const;
```
arsenius7: Maybe it needs to be renamed from `verbose` to `logLevel` now.
Also, as discussed verbally, another option is an environment variable controlling log verbosity for all `MorphIO` methods.

src/mut/section.cpp
```diff
@@ -203,57 +203,23 @@ std::shared_ptr<Section> Section::appendSection(
...
-    return !_compare(other, false);
+    return diff(*this, other, LogLevel::ERROR);
 }
 
 } // end namespace mut
 } // end namespace morphio
+
+std::ostream& operator<<(std::ostream& os, const morphio::mut::Section& section)
+{
+    auto points = section.points();
```
arsenius7: I guess this can not be empty, right?

src/properties.cpp
```diff
@@ -80,19 +49,19 @@ PointLevel& PointLevel::operator=(const PointLevel& other)
...
 bool compare(const std::vector<T>& vec1, const std::vector<T>& vec2,
-    const std::string& name, bool verbose_)
+    const std::string& name, LogLevel verbose)
 {
     if (vec1 == vec2)
         return true;
 
     if (vec1.size() != vec2.size()) {
-        if (verbose_)
+        if (verbose > LogLevel::ERROR)
```
arsenius7: with global log level we can move it to LBERROR macro

PR #77, opened 2019-06-04
Get version using git tags

wizmer:


wizmer:
For hybrid packages we cannot only rely on setuptools_scm as only the C++ might be compiled. This means CMake must be able to get the version number by itself.

This first proposal here relies on a call to `git describe --tags --dirty` in the CMakeLists to get the version that will be returned when morphio.version is called. However, the python package distribution version will be returned by the setuptools_scm package.

This is not perfect since on non tagged commit the distribution name and the version will differ. Example, let's say last tag was v0.1.1:
distribution: 0.1.2.devX-{some hash}
version with the package: 0.1.1-dirty

On tagged package, the version should be the same though.

arsenius7:
Seems like a reasonable compromise to me.

I tried to find some "stock" CMake helper module for extracting version from tags some time ago, but back then I have only found [this](https://github.com/rpavlik/cmake-modules/blob/master/GetGitRevisionDescription.cmake).

Until it proves to be not enough, I guess a single shell call would suffice.
Maybe `sh` and not `bash`, if it makes any difference?

You might also want to sync up with Tristan on that.
I have a feeling that he is extracting version using Python for Basalt:
https://github.com/tristan0x/basalt/blob/master/CMakeLists.txt

arsenius7:
Given that we already have `setup.py` in repo root and run tests through Python as well, the second approach looks fine to me.

I think for stuff like that we should get +2 from all reviewers. ^^

wizmer:
Fine. I'll wait for @mgeplf 's input

VERSION
```diff
@@ -1 +0,0 @@
-2.1.1
```
arsenius7: probably you still declare `VERSION` file in MANIFEST.in

src/CMakeLists.txt
```diff
@@ -29,7 +29,13 @@ set(MORPHIO_SOURCES
   mut/writers.cpp
   mut/modifiers.cpp)
 
-file(STRINGS ${CMAKE_CURRENT_SOURCE_DIR}/../VERSION MORPHIO_VERSION_STRING)
+execute_process(COMMAND python setup.py --version
```
mgeplf: debugging this will be interesting when it fails, if I understand this right:
cmake calls into setup.py, but it may already be running from setup.py in a bunch of sub-processes because it's called from setup.py

have you tested that this works from clean repository pull and such? I'm concerned that since the version is calculated from git, when working on this in the 'editable' manor, or as pip/setuptools changes, (like when it copies all the files required for building to a different folder) that we'll run into problems

Also, how does the versioning interact w/ spack?

setup.py
```diff
@@ -99,5 +92,7 @@ def build_extension(self, ext, cmake):
               'neurolucida'
               'neuromorphology'),
     zip_safe=False,
-    classifiers=[]
+    classifiers=[],
+    use_scm_version=True,
+    setup_requires=['setuptools_scm'],
```
mgeplf: hmm, I find the issue list for setuptools_scm a bit concerning - quite a lot of old issues, and it seems if it's not being used in the *exact* manor, it's hard to work around (ie: issue 304)

arsenius7: On the other hand, the repo is hosted in `pypa` and seems to be alive so far.
Also, I don't know if we ever need anything exotic from it. ^^

What are our alternatives for pairing version as seen by C++ / Python / Git tags?

src/CMakeLists.txt
arsenius7: > Also, how does the versioning interact w/ spack?

Either we use the same Git tag for fetching source from the repo; or fetch source tarball from PyPI as in https://github.com/BlueBrain/spack/pull/372.

wizmer: Yes, I just tested, it seems to work when installing in editable mode or from the tarball.

arsenius7:
Should we automatically upload source tarball to PyPI on tagging?
(https://docs.travis-ci.com/user/deployment/pypi/#deploying-tags ?)

wizmer:
I'd say it does not hurt.

mgeplf:
Ok, so basically we're going to trial using setuptools_scm in this repo?  I can handle that.

@wizmer: can you just answer https://github.com/BlueBrain/MorphIO/pull/77/files#r290259093 ?

setup.py
```diff
@@ -60,9 +60,7 @@ def build_extension(self, ext, cmake):
             build_args += ['--', '-j']
 
         env = os.environ.copy()
-        env['CXXFLAGS'] = '{} -DVERSION_INFO=\\"{}\\"'.format(
-            env.get('CXXFLAGS', ''),
-            self.distribution.get_version())
+        env['CXXFLAGS'] = env.get('CXXFLAGS', '')
```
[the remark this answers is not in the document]
wizmer: No. Do you want to remove this line ?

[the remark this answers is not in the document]
mgeplf: if it's not necessary, probably best to remove so as to not be confusing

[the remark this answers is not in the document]
wizmer: I was wondering if there could be some use cases where it could help people compile morphio. Like maybe passing an external include directory or disabling an error ?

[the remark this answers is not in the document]
mgeplf: sorry, wasn't clear, I mean specifically the `env['CXXFLAGS'] = env.get('CXXFLAGS', '')` line; if CXXFLAGS already exists, it's passed through, but if it doesn't a blank one is created, no?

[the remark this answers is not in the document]
wizmer: I see. Then, we don't need `env` variable at all it seems.

mgeplf:
lgtm

PR #79, opened 2019-06-19
Return all sections.

matz-e:
Somehow the first of the branches was skipped when returning sections from immutable morphologies.

### DIFF
--- a/src/morphology.cpp
+++ b/src/morphology.cpp
@@ -129,9 +129,10 @@ const std::vector<Section> Morphology::sections() const
 {
     // TODO: Make this more performant when needed
     std::vector<Section> sections_;
-    for (uint i = 0; i < _properties->get<morphio::Property::Section>().size();
-         ++i) {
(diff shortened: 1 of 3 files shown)


wizmer:
Wow, thanks for spotting this. I hope that won't have too many consequences.

src/morphology.cpp
```diff
@@ -129,7 +129,7 @@ const std::vector<Section> Morphology::sections() const
 {
     // TODO: Make this more performant when needed
     std::vector<Section> sections_;
```
tristan0x: What about using `std::vector::reserve` before filling `sections_`?
It would also use `emplace_back` to avoid the copies (including the one in `section` member method).

```
sections_.emplace_back(id, _properties);
```

matz-e: Good catch!

mgeplf: not sure there is a move ctr for Section, something to look into/add

PR #83, opened 2019-06-26
Fix annototations not being copied in morphio.mut.Morphology ctor

wizmer:


tests/test_5_mut.py
```diff
@@ -426,3 +427,24 @@ def test_non_C_nparray():
...
+            with tmp_asc_file('''((Dendrite)
+                      (3 -4 0 2)
+                      (3 -6 0 2)
+                      (3 -8 0 2)
+                      (3 -10 0 2)
+                      (
+                        (3 -10 0 2)
+                        (0 -10 0 2)
+                        (-3 -10 0 2)
+                        |       ; <-- empty sibling but still works !
```
mgeplf: is the annotation?

wizmer: No, no. The annotation is just that there is a single child. This is just an additionnal comment.

PR #86, opened 2019-07-11
Remove morphio::Section::length()

wizmer:
1) It was confusing because it was euclidian length and not pathlength
2) It was not even exposed in the binding
3) It will be in the higher-level library

And fix a few docstrings

### DIFF
--- a/binds/python/bind_misc.cpp
+++ b/binds/python/bind_misc.cpp
@@ -218,13 +218,13 @@ static void bind_misc(py::module &m) {
 
 
   m.def("diff", static_cast<bool (*)(const morphio::Morphology& left, const morphio::Morphology& right, morphio::enums::LogLevel logLevel)>(&morphio::diff),
-        "Perform a diff on 2 morphologies", "left"_a, "right"_a, "log_level"_a=morphio::enums::LogLevel::INFO);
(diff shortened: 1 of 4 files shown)


matz-e:
> However, I'm not convinced that `length()` should be removed. I'm open to the idea, but it feels like such a fundamental property (possibly the only one? no ambiguity, etc) of a section that it seems worth keeping?
> 
> I definitely think that the euclidean distance version of it should be removed.

Anything other than the euclidean distance will require additional calculations (whose results could be cached). I thought that was out of scope? `length` as such is also ambiguous. I'm wondering if the euclidean has many use(r)s?

wizmer:
I'm okay for putting the section distance along the arc instead (aka pathlength).
Should I name it length or pathlength though.
I don't like `length` because it is ambiguous with the euclidian distance.
I don't like `pathlength` because it is ambiguous with the distance to soma.
Maybe I think too hard...

wizmer:
>  I'm wondering if the euclidean has many use(r)s?
Yes. That's why I'm removing it.

wizmer:
But overall, I agree with Matthias, I would prefer add the pathlength in the higher level lib.

PR #91, opened 2019-07-19
Remove all notions of morphology equality

wizmer:
This is now handled in morph-tool


tests/test_5_mut.py
```diff
@@ -310,77 +310,6 @@ def test_mitochondria():
     assert_array_equal(np.array(first_child.relative_path_lengths, dtype=np.float32),
                        np.array([0.6, 0.7, 0.8, 0.9], dtype=np.float32))
 
-def test_equality():
```
wizmer: These tests have been replicated in morph-tool

mgeplf:
I think the version number should change, because this is a big change in functionality (in the right direction, IMO)

wizmer:
Ok, I'll tag this commit with an increment in the minor version number

PR #93, opened 2019-07-25
Fix case where SWC has a bifurcation at root point

wizmer:
When there was a bifurcation at root point, one neurite would go missing

### DIFF
--- a/tests/data/reversed_NRN_neurite_order.swc
+++ b/tests/data/reversed_NRN_neurite_order.swc
@@ -2,5 +2,8 @@
 # In this file the order is reversed
  1 1  0  0 0 1. -1
  2 4  1  1 1 1.  1 # apical
- 3 3  2  2 2 1.  1 # basal
- 4 2  3  3 3 1.  1 # axon
+ 3 4  2  1 1 1.  2
+ 4 3  0  1 2 1.  1 # basal
+ 5 3  0  1 3 1.  4
+ 6 2  3  3 3 1.  1 # axon
+ 7 2  3  3 4 1.  6
--- a/src/plugin/morphologySWC.cpp
+++ b/src/plugin/morphologySWC.cpp
@@ -163,7 +163,7 @@ public:
         return (sample.parentId == SWC_UNDEFINED_PARENT && sample.type != SECTION_SOMA);
     }
 
-    inline bool isRootSection(const Sample& sample)
+    inline bool isRootPoint(const Sample& sample)
     {
         return isOrphanNeurite(sample) ||
                (sample.type != SECTION_SOMA &&
@@ -172,7 +172,7 @@ public:
 
     inline bool isSectionStart(const Sample& sample)
     {
-        return (isRootSection(sample) || (sample.parentId > -1 && isSectionEnd(samples[static_cast<unsigned int>(sample.parentId)]))); // Standard section
+        return (isRootPoint(sample) || (sample.parentId > -1 && isSectionEnd(samples[static_cast<unsigned int>(sample.parentId)]))); // Standard section
     }
 
     inline bool isSectionEnd(const Sample& sample)
@@ -284,6 +284,12 @@ public:
         _pushChildren(depthFirstSamples, -1);
         for (const auto id : depthFirstSamples) {
             const Sample& sample = samples[id];
+
+            // Bifurcation right at the start
+            if(isRootPoint(sample) && isSectionEnd(sample)) {
+                continue;
+            }
+
             if (isSectionStart(sample)) {
                 _processSectionStart(sample);
             } else if (sample.type != SECTION_SOMA) {
@@ -324,7 +330,7 @@ public:
 
         std::shared_ptr<morphio::mut::Section> section;
 
-        if (isRootSection(sample)) {
+        if (isRootPoint(sample)) {
             section = morph.appendRootSection(properties, sample.type);
         } else {
             // Duplicating last point of previous section if there is not already a duplicate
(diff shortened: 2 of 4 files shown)


tests/data/reversed_NRN_neurite_order.swc
```diff
@@ -2,5 +2,8 @@
 # In this file the order is reversed
  1 1  0  0 0 1. -1
  2 4  1  1 1 1.  1 # apical
- 3 3  2  2 2 1.  1 # basal
```
wizmer: Those were single point neurites so they were not doing any sense.

mgeplf: Aren't they?  I thought SWC says that they connect to line 1, no?

wizmer: Huh, if you have a link that says so I'll take it.
I thought we had agreed that the first point of the neurite was the first point of the neurite. And not the soma point it connects to.

mgeplf: > I thought we had agreed that the first point of the neurite was the first point of the neurite. 

That's for ASCII files.

For SWC:  `Connectivity is expressed with this value.`: http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html

I think we should get @wvangeit's opinion.

wvangeit: I'd say a single point neurite does make sense. But how would the surface area be interpreted ? Because if it means the first point is one with a diameter of the soma, that would not really be correct either.

wizmer: We already discussed that and said that auto-connecting to a soma point was too dangerous because it is ambiguous.

Most of the time, it would even be plain wrong. Like for example in the case of a three-point soma representation it is clearly stated in http://neuromorpho.org/SomaFormat.html :
```
The second and third soma points, as well as all starting points (roots) of dendritic and axonal arbors have this first point as the parent (parent ID 1).
```
In this case, if the soma point is used as the first point of the neurite, it means all neurites will start at the soma center, which is totally off if the radius is large.

For me it would be safer to forbid one point root sections and to let people remove the ambiguity themselve. If they want to use the soma point as the first point of the section, they can just add this point in the file.

wvangeit: I agree that it's complicated and that we have to see how this can be solved. (could we add a point at the intersection with the soma ?).

> For me it would be safer to forbid one point root sections and to let people remove the 
> ambiguity themselve. If they want to use the soma point as the first point of the section, 
> they can just add this point in the file.
But this easier said than done. If somebody want to analyse a lot of morphologies from existing databases, and these morphologies have this 'problem' I'm not sure we should force people to 'fix' these morphologies, since afaik these single point neurites don't contradict the SWC format description.

wizmer: Just to be clear: the only thing I want to forbid is a one point section at the root, without children. I don't think there are many morphologies that would fall under this category.

Yes we could add a point at the intersection of the soma, but this is just an heuristic. And in fact, it would hurt more often than not. 
When an experimentalist decides to click somewhere on the image to place the first point of the neurite, most of the time, he has a good reason to do so. He has the opportunity to put this first point on the soma surface if we wants to, if he does not I guess he has a good reason to. Adding an euristic to MorphIO is like discarding its judgement.

Here is the first file I got when I took a random file on neuromorpho.org:
http://neuromorpho.org/dableFiles/bilkei-gorzo/Source-Version/wtlpsF3254Iba2_4.swc

Let's look at the X coordinate.
Along the neurite, the resolution is pretty high, the experimentalist put a point every ~0.5 micron even though the first point of the neurite is 30 micron of the soma center. The experimentalist probably had a good reason to do this. 
Second neurite starts at ID17. This time the experimentalists made the neurite *exactly* start at the soma center, this is explicit, he really clicked here on purpose. If he wanted to do the same with the first neurite, he would have done it.

```
1 1 36.43578643578644 96.13997113997114 4.001621333333333 0.36117243867243864 -1
2 1 62.59018759018759 101.01010101010101 2.0008106666666663 0.38472041847041855 1
3 1 63.67243867243867 100.82972582972583 2.0008106666666663 0.37116522366522364 2
4 1 64.39393939393939 100.46897546897547 2.0008106666666663 0.4158694083694084 3
5 1 64.93506493506494 100.46897546897547 1.5006079999999997 0.3877958152958153 4
6 1 65.83694083694084 99.20634920634922 1.5006079999999997 0.31576479076479075 5
...
16 1 71.06782106782107 98.12409812409813 2.501013333333333 0.338495670995671 15
17 3 36.43578643578644 96.13997113997114 4.001621333333333 0.36117243867243864 1
18 3 25.432900432900436 117.24386724386724 2.0008106666666663 0.3708820346320346 17
19 3 25.432900432900436 119.76911976911977 3.0012159999999994 0.36074675324675326 18
...
```

mgeplf: > Just to be clear: the only thing I want to forbid is a one point section at the root

Unfortunately, the 'ball and stick' is pretty much the go to example of a morphology when people are learning about simulations, from what I can tell (ex: https://www.neuron.yale.edu/neuron/static/docs/neuronpython/ballandstick1.html)

wvangeit: I guess we'll have to see how many times this happens in the wild, and might have to revisit if it happens too often

mgeplf: BTW, I tried loading both of these SWC files in nrngui, and they failed for some reason...sadly, it said the 'error was on the console'; but there was nothing there (except a missing font).

@wvangeit can you try?

wizmer: > > Just to be clear: the only thing I want to forbid is a one point section at the root
> 
> Unfortunately, the 'ball and stick' is pretty much the go to example of a morphology when people are learning about simulations, from what I can tell (ex: https://www.neuron.yale.edu/neuron/static/docs/neuronpython/ballandstick1.html)

You can do a ball and stick. You just specify the start of the neurite. 
```
1 1 0 0 0 1 -1
2 3 0 0 0 1  1
3 3 1 0 0 1  2
```
Just don't:
```
1 1 0 0 0 1 -1
1 3 1 0 0 1  1
```

PR #95, opened 2019-08-05
feature: implement max_distance

### LINKED ISSUE #94: Missing documentation and implementation for Soma::maxDistance
Center of gravity is already defined. Could this be the maximum distance from the center of gravity? We rely on this, would be nice to have it implemented.

https://github.com/BlueBrain/MorphIO/blob/ee04bbed836490c2ee7882d63a596f388c782ee5/include/morphio/soma.h#L66-L67

https://github.com/BlueBrain/MorphIO/blob/ee04bbed836490c2ee7882d63a596f388c782ee5/src/soma.cpp#L43-L46

matz-e:
Fixes #94.

### DIFF
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -122,6 +122,9 @@ static void bind_immutable_module(py::module &m) {
                                              return py::array(3, soma->center().data());
                                          },
             "Returns the center of gravity of the soma points")
+        .def_property_readonly("max_distance", &morphio::Soma::maxDistance,
+                               "Return the maximum distance between the center of gravity "
+                               "and any of the soma points")
         .def_property_readonly("type",
                                &morphio::Soma::type,
                                "Returns the soma type")
--- a/src/mut/soma.cpp
+++ b/src/mut/soma.cpp
@@ -38,6 +38,23 @@ float Soma::surface() const
         points());
 }
 
+float Soma::maxDistance() const
+{
+    const auto ps = points();
+    if (ps.size() == 0) {
+        return 0.;
+    }
+    const auto c = center();
+    float dist = distance(c, ps[0]);
+    for (size_t i = 1; i < ps.size(); ++i) {
+        const float d = distance(c, ps[i]);
+        if (d > dist) {
+            dist = d;
+        }
+    }
+    return dist;
+}
+
 std::ostream& operator<<(std::ostream& os, Soma& soma)
 {
     os << dumpPoints(soma.points());
--- a/src/soma.cpp
+++ b/src/soma.cpp
@@ -42,7 +42,19 @@ float Soma::surface() const
 
 float Soma::maxDistance() const
 {
-    return 0;
+    const auto ps = points();
+    if (ps.size() == 0) {
+        return 0.;
+    }
+    const auto c = center();
+    float dist = distance(c, ps[0]);
+    for (size_t i = 1; i < ps.size(); ++i) {
+        const float d = distance(c, ps[i]);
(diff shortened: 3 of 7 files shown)


matz-e:
Any opinions on this? Or should it be moved to MorphoKit?

wizmer:
I'll add it to morpho-kit instead if you don't mind.

wizmer:
My bad, I changed my mind. It makes sense to put it here in fact.

wizmer:
Thanks for implementing the feature.

src/mut/soma.cpp
```diff
@@ -38,6 +38,23 @@ float Soma::surface() const
         points());
 }
 
+float Soma::maxDistance() const
+{
+    const auto ps = points();
+    if (ps.size() == 0) {
```
mgeplf: I wonder if we can do something like:
```
    const auto c = center();
    return std::accumulate(begin(ps), end(ps),
        0.f,
        [](T a, T b){return std::max(a, distance(c, b));});
```

src/soma.cpp
```diff
@@ -42,7 +42,19 @@ float Soma::surface() const
 
 float Soma::maxDistance() const
```
mgeplf: is there a place where duplicate functionality can be called put, so we don't have to maintain two versions?  kind of like how centerOfGravity works

binds/python/bind_immutable.cpp
```diff
@@ -122,6 +122,9 @@ static void bind_immutable_module(py::module &m) {
                                              return py::array(3, soma->center().data());
                                          },
             "Returns the center of gravity of the soma points")
+        .def_property_readonly("max_distance", &morphio::Soma::maxDistance,
+                               "Return the maximum distance between the center of gravity "
+                               "and any of the soma points")
```
mgeplf: This makes sense for contour style somas, but for cylinder ones, it satisfies the docstring, but I'm wondering if the radius of the cylinders is accounted for.

This also applies to the centerOfGravity calculation - so maybe that should be a separate issue

src/mut/soma.cpp
tristan0x: +1. I would use `{}` instead of `0.f` though

tristan0x: Also I would use `ps.empty()` instead of `ps.size() == 0`.

mgeplf: Yeah, {} is more generic.
AFAIK, the ps.size() check isn't needed w/ the above code - if std::accumulate is given an empty itr, it returns the initial value.

I should have been more clear:
```

float Soma::maxDistance() const
{
    const auto ps = points();
    const auto c = center();
    return std::accumulate(begin(ps), end(ps),
        {},
        [](float a, float b){return std::max(a, distance(c, b));});
}
```

matz-e: Sounds like an abuse of std::accumulate, but should work. I was thinking about `std::max_element` before, and that calculates too much.

binds/python/bind_immutable.cpp
matz-e: By now, it should be separate — handling should be solved for both `maxDistance` and `centreOfGravity` at the same time, /me thinks.

src/soma.cpp
matz-e: Sure. I can follow up.

src/mut/soma.cpp
mgeplf: since the std library doesn't have `reduce` until c++17, std::accumulate has been used instead for a long time - so it's a 'standard' usage /grumpy old man

PR #96, opened 2019-08-08
Deduplicate and simplify code.

matz-e:
As suggested in #95.

Apparently, I'm unable to fix the LLVM compilation. Any ideas?

### DIFF
--- a/src/mut/soma.cpp
+++ b/src/mut/soma.cpp
@@ -40,7 +40,7 @@ float Soma::surface() const
 
 float Soma::maxDistance() const
 {
-    return maxDistanceToCenterOfGravity(_properties->_somaLevel._points);
+    return maxDistanceToCenterOfGravity(_pointProperties._points);
 }
 
 std::ostream& operator<<(std::ostream& os, Soma& soma)
--- a/src/mut/soma.cpp
+++ b/src/mut/soma.cpp
@@ -40,7 +40,7 @@ float Soma::surface() const
 
 float Soma::maxDistance() const
 {
-    return maxDistanceToCenterOfGravity(points());
+    return maxDistanceToCenterOfGravity(_properties->_somaLevel._points);
 }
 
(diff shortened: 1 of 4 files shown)


wizmer:
I think there is a constness issue.
It wants to instantiate `'morphio::maxDistanceToCenterOfGravity<gsl::span<const std::array<float, 3> > >`
although you explicitely instantiated:
template float maxDistanceToCenterOfGravity(const range<const Point>& points);

src/mut/soma.cpp
```diff
@@ -40,19 +40,7 @@ float Soma::surface() const
...
-    const auto c = center();
-    float dist = distance(c, ps[0]);
-    for (size_t i = 1; i < ps.size(); ++i) {
-        const float d = distance(c, ps[i]);
-        if (d > dist) {
-            dist = d;
-        }
-    }
-    return dist;
+    return maxDistanceToCenterOfGravity(points());
```
wizmer:
```suggestion
    return maxDistanceToCenterOfGravity(_properties->_somaLevel._points);
```

wizmer: But `points()` should already be a `const range` so I don't understand.

src/mut/soma.cpp
```diff
@@ -40,19 +40,7 @@ float Soma::surface() const
...
-    const auto c = center();
-    float dist = distance(c, ps[0]);
-    for (size_t i = 1; i < ps.size(); ++i) {
-        const float d = distance(c, ps[i]);
-        if (d > dist) {
-            dist = d;
-        }
-    }
-    return dist;
+    return maxDistanceToCenterOfGravity(_properties->_somaLevel._points);
```
wizmer: Sorry, I wanted to suggest on the other one (the non mutable)

matz-e: Just got that.

PR #99, opened 2019-09-06
Iterators without templates

wizmer:
The compilation on mac would fail because of a forward declaration not resolved properly when defining the iterators. Removing the templatization of the iterators fixes it.

The error is the following: https://travis-ci.com/BlueBrain/MorphIO/jobs/209325891#L2037
Using Xcode: xcode10.1 and the compiler AppleClang 10.0.0.10001145

The class Section has a function `breadth_begin` that returns an iterator `breadth_iterator` which is an alias to `Iterator<std::queue<queue<Section>>>`. From what I could see, you can't forward declare a template class, ie. you can't write `class Iterator<std::queue<queue<Section>>>` in `section.h` so you have to import `iterators.h` to make it known.
Now, in `iterators.h` you need to forward declare `Section` (because if you import `section.h` you run into a circular dependency.
So far so good, but then you need to explicitly instantiate `Iterator<std::queue<queue<Section>>>` with `extern template class Iterator<std::queue<queue<Section>>>;`. And this is where the compilator complains.

```
  In file included from /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/pip-req-build-9FTZRy/src/iterators.cpp:1:

  In file included from /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/pip-req-build-9FTZRy/src/../include/morphio/iterators.h:3:

  In file included from /Applications/Xcode-10.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/queue:170:

  /Applications/Xcode-10.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/deque:268:34: error: invalid application of 'sizeof' to an incomplete type 'morphio::Section'
```


src/morphology.cpp
```diff
@@ -123,7 +123,7 @@ const std::vector<Section> Morphology::sections() const
     std::vector<Section> sections_;
     auto count = _properties->get<morphio::Property::Section>().size();
     sections_.reserve(count);
-    for (uint i = 0; i < count; ++i) {
+    for (unsigned int i = 0; i < count; ++i) {
```
mgeplf: is count size_t?  if so, why not use that?

include/morphio/iterators.h
```diff
@@ -52,31 +29,5 @@ class graph_iterator
...
-template <>
-Section upstream_iterator::operator*() const;
-template <>
-depth_iterator& depth_iterator::operator++();
-template <>
-breadth_iterator& breadth_iterator::operator++();
-template <>
-upstream_iterator& upstream_iterator::operator++();
-
-extern template class Iterator<std::stack<Section>>;
```
mgeplf: no longer need #include <queue>

Confused as to why CI doesn't pick up such things, I just happened to check.

include/morphio/iterators/breadth.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
mgeplf: If I'm reading this right, you basically just manually templated things - which seems counter-intuitive to me.

I think it would be worthwhile figuring out why the initial code didn't work.

src/CMakeLists.txt
```diff
@@ -18,10 +17,16 @@ set(MORPHIO_SOURCES
   vasc/properties.cpp
   vector_utils.cpp
   version.cpp
+	iterators/upstream.cpp
```
mgeplf: why the indentation?

wizmer: because visual studio tabs...

src/morphology.cpp
wizmer: Because then you need to cast in back to unsigned int when you use `i` in the Section constructor.

include/morphio/iterators/breadth.h
wizmer: I have updated the diagnostic. I'm pretty sure there is a way to solve this without getting rid of the templates but I already spent a lot of time on it and this is blocking the compilation on mac so this is quite a big issue.

mgeplf:
Can you repro this on your own mac? I think there must be a better way.

wizmer:
If by this you mean the non-working behaviour, yes I can. I'll bring my mac tomorrow.

mgeplf:
I spent some time on this; seems to be a `libc++` problem that can be repro'd w/:
```
#include <queue>

struct foo {
  std::queue<foo> q;
};
```

wizmer:
Replaced by https://github.com/BlueBrain/MorphIO/pull/112

PR #101, opened 2019-09-11
enumeration for VascularSectionType

SnigdhaD:


tests/test_4_immut.py
```diff
@@ -126,6 +126,7 @@ def test_components_vasculature():
     assert_equal(len(morphology.section_types), 3080)
     assert_equal(len(morphology.section(0).predecessors), 0)
     assert_equal(len(morphology.section(0).successors), 2)
+    assert_equal(morphology.section(0).type, 0)
```
wizmer: 1) Compare to the actual type, not the int
2) the type is SECTION_NOT_DEFINED, can you change the type in the h5 file to have a more meaningful one.
3) Can you create another test file and put an undefined type (like let's say 12), and create a test with it ?

tests/test_4_immut.py
```diff
@@ -126,6 +126,14 @@ def test_components_vasculature():
     assert_equal(len(morphology.section_types), 3080)
     assert_equal(len(morphology.section(0).predecessors), 0)
     assert_equal(len(morphology.section(0).successors), 2)
+
```
wizmer: This test function is becoming to big.
I think it's time to make a test_vasculature.py and create smaller test function there.

SnigdhaD: I agree. So I will move all the vasculature tests to the new file.

tests/test_4_immut.py
wizmer: How did you test the neuron type. I just attempted to read a neuron with wrong section type and got the following error message:
```
RawDataError: 
tests/data/simple.swc:25:error
Unsupported section type: 13
```
I think vasculature should print a similar message.

SnigdhaD: Alright.
I checked with the simple.h5 file (changed the type of one of the sections manually)

SnigdhaD: I think this is because the IO files for H5 for both neuron and vasculature do not check the section type at the time of reading - https://github.com/BlueBrain/MorphIO/blob/master/src/plugin/morphologyHDF5.cpp

wizmer: Can you add the check in both files and LBTROW ErrorMessages::ERROR_UNSUPPORTED_SECTION_TYPE ?

include/morphio/enums.h
```diff
@@ -151,7 +151,8 @@ enum VascularSectionType
     SECTION_ARTERIOLE = 4,
     SECTION_VENOUS_CAPILLARY = 5,
     SECTION_ARTERIAL_CAPILLARY = 6,
-    SECTION_TRANSITIONAL = 7
+    SECTION_TRANSITIONAL = 7,
+    SECTION_CUST_START = 8
```
wizmer: CUSTOM

SnigdhaD: that clashes with the enum in SECTION_TYPE

eleftherioszisis: CUSTARD?

wizmer: https://stackoverflow.com/a/23288968/2533394

tests/test_10_vasculature.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def test_iterators_vasculature():
+    morphology = vasculature.Vasculature(os.path.join(_path, "h5/vasculature1.h5"))
+    assert_array_equal([sec.id for sec in morphology.sections], range(3080))
+    assert_equal(len([section.id for section in morphology.iter()]), 3080)
+    all_sections = set([sec.id for sec in morphology.sections])
+    for sec in morphology.iter():
+        all_sections.remove(sec.id)
+    assert_equal(len(all_sections), 0)
```
wizmer: always add newline at end of file

PR #102, opened 2019-09-11
versioning: rely on git.

matz-e:


src/CMakeLists.txt
```diff
@@ -28,12 +28,14 @@ set(MORPHIO_SOURCES
...
-                ERROR_VARIABLE PY_VERSION_ERROR
-                OUTPUT_STRIP_TRAILING_WHITESPACE)
+if(NOT MORPHIO_VERSION_STRING)
+    execute_process(COMMAND git describe --tags
+                    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+                    RESULT_VARIABLE GIT_VERSION_FAILED
+                    OUTPUT_VARIABLE MORPHIO_VERSION_STRING
+                    ERROR_VARIABLE GIT_VERSION_ERROR
+                    OUTPUT_STRIP_TRAILING_WHITESPACE)
+endif()
```
wizmer: Here you don't need the:
```
 # keep last line of command output 
   string(REPLACE "\n"
               ";"
                   ";"
               PY_PKG_VERSION_FULL
                   GIT_PKG_VERSION_FULL
               "${PY_PKG_VERSION_FULL}")
                   "${GIT_PKG_VERSION_FULL}")
list(GET PY_PKG_VERSION_FULL -1 PY_PKG_VERSION_FULL)
   list(GET GIT_PKG_VERSION_FULL -1 SONATA_VERSION)
```
?

matz-e: We didn't before, either :)

We need that snippet when telling CMake about the version information. How this project is set up, the version number we extract from git/get passed from `setup.py`/extracted from `setup.py`, the version number ends up in a string in the file you configure below.

I think this is actually better, since it retains more information. This is what I see when running `spack diy` in `version.cpp`:
```
#include <string>

#include <morphio/types.h>
namespace morphio
{
  const std::string VERSION("2.2.1.dev4+g4e5982e.d20190815");
}
```
Seems good, no?

PR #104, opened 2019-09-12
cleanup morphio/types.h morphio/enums.h

mgeplf:
* removed headers that aren't used in morphio/types.h
* remove LBWARN, use std::cerr
* removed cruft
* move ostream to src/enums.ccp


mgeplf:
the build fails, so I'm not ready to merge this.  I'll request a review when I need it.

PR #105, opened 2019-09-13
move graph_iterator to vasculature namespace

mgeplf:
* cleanup includes
* use Point & Points type alias where appropriate

### DIFF
--- a/src/iterators.cpp
+++ b/src/iterators.cpp
@@ -1,6 +1,5 @@
 #include <morphio/iterators.h>
 #include <morphio/section.h>
-#include <morphio/vasc/section.h>
 
 namespace morphio {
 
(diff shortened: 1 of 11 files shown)


src/iterators.cpp
```diff
@@ -170,6 +128,42 @@ upstream_iterator& upstream_iterator::operator++()
 }
 
 
+// Instantiations
```
wizmer: Do you think it could be worthwhile to put this chunk in a vasculature_iterator file ?

mgeplf: yeah, I think all the vasculature related things should be in vasculature; I'm not sure why they were in `types.h` in first place

mgeplf:
ping

PR #110, opened 2019-09-20
make iterators have less friend classes

mgeplf:


mgeplf:
I think we don't need many of the friends - the compilation 'test' passed, but maybe it's for interface stuff?

mgeplf:
ping

mgeplf:
Why are were the friends needed?

wizmer:
Because at the beginning I did not want to expose the default constructor. Like when you removed the friend from mito_iterator you had to move MitoIterator() to public.

mgeplf:
Why not allow the default ctor?  Anyways, the fact that these classes are diverging is troublesome.

PR #111, opened 2019-09-24
cleanup readers

mgeplf:
* put into 'readers' namespace, plugins was a brion thing, and it means  something different
* remove unused details code, moved used code to readers (only thing  that was using it)


binds/python/bind_misc.cpp
```diff
@@ -227,6 +227,4 @@ static void bind_misc(py::module &m) {
         .def(py::init<std::vector<uint32_t>, std::vector<float>,
              std::vector<morphio::Property::Diameter::Type>>(),
              "neuronal_section_ids"_a, "distances_to_section_start"_a, "diameters"_a);
-
-    m.doc() = "pybind11 example plugin"; // optional module docstring
```
mgeplf: not needed?

src/readers/morphologyHDF5.cpp
```diff
@@ -21,15 +20,38 @@
...
-#include "../detail/lockHDF5.h"
-#include "../detail/morphologyHDF5.h"
-#include "../detail/utilsHDF5.h"
+#include "utilsHDF5.h"
+
+#include <highfive/H5Utility.hpp> // HighFive::SilenceHDF5
+
+namespace {
+// v1 & v2
+const std::string _d_points("points");
```
mgeplf: all these static ctors are a bit annoying, imo

src/readers/morphologySWC.cpp
```diff
@@ -328,32 +339,33 @@ class SWCBuilder
     {
         Property::PointLevel properties;
 
-        std::shared_ptr<morphio::mut::Section> section;
+        uint32_t id = 0;
```
mgeplf: there should really be an ID type, I think

PR #112, opened 2019-09-25
rework iteration

mgeplf:
* have a single template for each of {upstrea,breadth,depth}_iterator_t
* don't have duplicated code for section/mito_section/mut_section/mut_mitosection
* change breadth_iterator of Morphology to iterate outwards from soma to
  each neurite equally, instead of doing a breadth_first on each neurite
* enable clang on osx building


include/morphio/morphology.h
```diff
@@ -99,17 +100,17 @@ class Morphology
...
        If id == -1, the iteration will be successively performed starting
        at each root section
     **/
-    breadth_iterator breadth_begin() const;
-    breadth_iterator breadth_end() const;
+    breadth_iterator_t<Section, Morphology> breadth_begin() const;
+    breadth_iterator_t<Section, Morphology> breadth_end() const;
 
     /**
      * Return the soma type
```
wizmer: Can you nuke the commented function `traverse` a bit below in this file ?

include/morphio/section_iterators.hpp
```diff
@@ -0,0 +1,252 @@
...
+
+    SectionT operator*() const
+    {
+        return deque_.front();
+    }
+
+    breadth_iterator_t operator++()
+    {
+        if (deque_.empty()) {
+            LBTHROW(MissingParentError("XXXXXXXXXXXXXXXXXXX"));
```
wizmer: It's not really a missing parent error no ?

include/morphio/morphology.h
wizmer: Can you nuke the commented function `traverse` a bit below in this file ? Thanks

include/morphio/section_iterators.hpp
```diff
@@ -0,0 +1,252 @@
...
+
+private:
+    std::deque<SectionT> deque_;
+};
+
+template<typename SectionT>
+class upstream_iterator_t {
+public:
+    upstream_iterator_t()
+    : unused(0)
```
wizmer: Explain the trick

include/morphio/section_iterators.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    SectionT operator*() const
+    {
+        return current;
+    }
+
+    upstream_iterator_t operator++()
+    {
+        //int a = *((int*) 0);
```
wizmer: remove comment

.travis.yml
```diff
@@ -12,6 +12,13 @@ matrix:
       env:
         CXX: clang++
         CC: clang
+    - os: osx
+      osx_image: xcode10.1
+      language: generic
+      python: 3.6
+      env:
+        CXX: clang++
+        CC: clang
```
wizmer: Should we add the osx gcc build also ?

include/morphio/morphology.h
```diff
@@ -99,17 +100,17 @@ class Morphology
 
        If id == -1, the iteration will start at each root section, successively
     **/
-    depth_iterator depth_begin() const;
-    depth_iterator depth_end() const;
+    depth_iterator_t<Section, Morphology> depth_begin() const;
```
wizmer: I suppose the old denomination `depth_iterator it = morphology.depth_begin()` still works ?

include/morphio/section_iterators.hpp
mgeplf: Nope, wanted to ask: which exception do you prefer?

wizmer: The basic `MorphioError` or a new `IteratorError` one ?

include/morphio/morphology.h
mgeplf: I don't follow

wizmer: My question is: is this change breaking the API or do we need to change morphokit as well ?

mgeplf: Ah, sorry, I see.  Could you try it?  Otherwise, I can make an type alias that should allow it to compile.  The ABI will be different, though, so we can't just link it.

wizmer: morphokit does not compile because of this.

```
/home/bcoste/workspace/morphology/morphokit/src/iterators.cpp:14:115: error: no matching function for call to ‘morphio::upstream_iterator_t<morphio::Section>::upstream_iterator_t(const morphio::Morphology&)’
                                  const Morphology* morphology) : BaseIterator(base_morpho), _morphology(morphology)
                                                                                                                   ^
In file included from /home/bcoste/.local/include/morphio/section.h:5:0,
                 from /home/bcoste/workspace/morphology/morphokit/src/../include/morphokit/iterators.h:2,
                 from /home/bcoste/workspace/morphology/morphokit/src/iterators.cpp:1:
/home/bcoste/.local/include/morphio/section_iterators.hpp:190:5: note: candidate: morphio::upstream_iterator_t<SectionT>::upstream_iterator_t(const morphio::upstream_iterator_t<SectionT>&) [with SectionT = morphio::Section]
     upstream_iterator_t(const upstream_iterator_t& other)
```

 The fix looks simple but that means the API is broken. However, I don't think we have other C++ users so we might not care.
Or would the `using depth_iterator = depth_iterator_t<....>` be possible ?

mgeplf: Hmm, did you try w/ the 3c424d7 version? I thought I added all the type aliases.

Also, it looks like it's the `upstream_iterator` that's failing, and I don't think it makes sense to upstream iterate a morphology, since there is no where to iterate to.  Is that call necessary?

wizmer: Yes, I tried on 3c424d7.

mgeplf: Ok, just making sure.

What about the point that upstream_iterator(Morphology) doesn't really make sense?

wizmer: That absolutely makes no sense, yes

wizmer: Deleting the upstream_iterator(Morphology) fixes the compilation.

PR #115, opened 2019-10-02
Implement endoplasmic reticulum

wizmer:
Provide an API to retrieve data in the /organelles/endoplasmic_reticulum
group.

One can retrieve the following properties:
- neuronal section indices
- volumes for each of the listed neuronal section
- surface ereas for each of the listed neuronal section
- filament counts for each of the listed neuronal section

Spec at:


wizmer:
This will be in another commit

include/morphio/endoplasmic_reticulum.h
```diff
@@ -0,0 +1,43 @@
...
+    **/
+    const std::vector<float>& surfaceAreas() const;
+
+    /**
+       Returns the number of filaments for each neuronal section
+    **/
+    const std::vector<uint32_t>& filamentCounts() const;
+
+private:
+    EndoplasmicReticulum(std::shared_ptr<Property::Properties> properties)
```
mgeplf: I'm still don't fully get why this is private, and then you have to declare `Morphology` a friend.  Seems like you're allowing `Morphology` much more access (ie: to _properties) than is necessary.

src/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ set(MORPHIO_SOURCES
     errorMessages.cpp
     mito_section.cpp
     mitochondria.cpp
+    endoplasmic_reticulum.cpp
```
mgeplf: please keep this sorted

src/endoplasmic_reticulum.cpp
```diff
@@ -0,0 +1,25 @@
+#include <morphio/properties.h>
+#include <morphio/endoplasmic_reticulum.h>
+
+namespace morphio {
+const std::vector<uint32_t>& EndoplasmicReticulum::sectionIndices() const
+{
+    return _properties->_endoplasmicReticulumLevel._sectionIndex;
+}
+
+const std::vector<float>& EndoplasmicReticulum::volumes() const
```
mgeplf: how is the lifetime of the vector guaranteed?

src/mitochondria.cpp
```diff
@@ -1,5 +1,6 @@
 #include <morphio/mito_section.h>
 #include <morphio/mitochondria.h>
+#include <morphio/properties.h>
```
mgeplf: why is this needed?

src/readers/morphologyHDF5.cpp
```diff
@@ -486,6 +494,29 @@ void MorphologyHDF5::_read(const std::string& groupName,
...
 }
 
+void MorphologyHDF5::_readEndoplasmicReticulum()
+{
+    {
+        HighFive::SilenceHDF5 silence;
+
+        try {
+            const auto group = _file->getGroup(_g_endoplasmic_reticulum);
+        } catch (HighFive::GroupException&) {
```
mgeplf: what's the semantics of getting this? can we check if the file has the property first, instead of catching the (potentially expensive) exception?

include/morphio/endoplasmic_reticulum.h
wizmer: > I'm still don't fully get why this is private, and then you have to declare Morphology a friend. 

My idea was that in MorphIO the *only* public constructor would be `Morphology`.
I don't want people to start creating EndoplasmicReticulum, or Section, Mitochondria objects that are not attached to a `Morphology` object.
You should only access those objects through `Morphology`:
```
m = Morphology("my-morph.swc");
m.mitochondria().do_stuff();
m.section(1).do_other_stuff();
m.endoplasmicReticulum().do_even_more_stuff();
```

> Seems like you're allowing Morphology much more access (ie: to _properties) than is necessary.

She only place where the EndoplasmicReticulum is called is in `Morphology::endoplasmicReticulum()` and the passed `_properties` structure is already owned by `Morphology`

Does this make sense ?

src/CMakeLists.txt
wizmer: good point

src/readers/morphologyHDF5.cpp
wizmer: This semantics comes straight from Brion. I can have a deeper look if you want.

mgeplf: Nope, it's ok.

tristan0x: There is a `NodeTraits::exist` method that does what you need:
https://github.com/BlueBrain/HighFive/blob/74250cf2f2a82a2b9af2f00f93475c431a2d7e09/include/highfive/bits/H5Node_traits.hpp#L110-L115

Otherwise I would consider:
```suggestion
        } catch (const HighFive::GroupException&) {
```

include/morphio/endoplasmic_reticulum.h
```diff
@@ -0,0 +1,43 @@
...
+        : _properties(properties)
```
tristan0x: To save an atomic increment in `_ptr` constructor and an atomic decrement in `properties` destructor:

```suggestion
        : _properties(std::move(properties))
```

wizmer: Good point. Thanks for the suggestion.

src/endoplasmic_reticulum.cpp
wizmer: The lifetime will be the one of the Morphology object.

src/mitochondria.cpp
wizmer: Not needed, indeed.

src/endoplasmic_reticulum.cpp
```diff
@@ -0,0 +1,25 @@
+#include <morphio/properties.h>
+#include <morphio/endoplasmic_reticulum.h>
+
+namespace morphio {
+const std::vector<uint32_t>& EndoplasmicReticulum::sectionIndices() const
```
tristan0x:
```suggestion
const std::vector<uint32_t>& EndoplasmicReticulum::sectionIndices() const noexcept
```

src/endoplasmic_reticulum.cpp
tristan0x:
```suggestion
const std::vector<float>& EndoplasmicReticulum::volumes() const noexcept
```

src/endoplasmic_reticulum.cpp
```diff
@@ -0,0 +1,25 @@
...
+{
+    return _properties->_endoplasmicReticulumLevel._sectionIndex;
+}
+
+const std::vector<float>& EndoplasmicReticulum::volumes() const
+{
+    return _properties->_endoplasmicReticulumLevel._volume;
+}
+
+const std::vector<float>& EndoplasmicReticulum::surfaceAreas() const
```
tristan0x:
```suggestion
const std::vector<float>& EndoplasmicReticulum::surfaceAreas() const noexcept
```

src/endoplasmic_reticulum.cpp
```diff
@@ -0,0 +1,25 @@
...
+{
+    return _properties->_endoplasmicReticulumLevel._surfaceArea;
+}
+
+const std::vector<uint32_t>& EndoplasmicReticulum::filamentCounts() const
```
tristan0x:
```suggestion
const std::vector<uint32_t>& EndoplasmicReticulum::filamentCounts() const noexcept
```

src/morphology.cpp
```diff
@@ -87,6 +88,11 @@ Mitochondria Morphology::mitochondria() const
     return Mitochondria(_properties);
 }
 
+const EndoplasmicReticulum Morphology::endoplasmicReticulum() const
```
tristan0x:
```suggestion
EndoplasmicReticulum Morphology::endoplasmicReticulum() const
```

src/readers/morphologyHDF5.cpp
```diff
@@ -479,6 +487,29 @@ void MorphologyHDF5::_read(const std::string& groupName,
...
 
+void MorphologyHDF5::_readEndoplasmicReticulum()
+{
+    {
+        HighFive::SilenceHDF5 silence;
+
+        try {
+            const auto group = _file->getGroup(_g_endoplasmic_reticulum);
+        } catch (const HighFive::GroupException&) {
+            return;
```
tristan0x: Don't we want to emit an error message here?

include/morphio/endoplasmic_reticulum.h
mgeplf: Ok, I think I get it

src/readers/morphologyHDF5.cpp
wizmer: No. The endoplasmic group is optional.

Issue #116, opened 2019-10-21
Starting point index of each section


WeinaJi:
Hello, 

Currently I am trying to inject the morphology objects from MorphIO API into our package SpatialIndex. The SpatialIndex API has a method to add branches with an array of points (start/end points of segments), an array of radii, and an array of the first point index of each branch. For the first two arrays, I can use the property accessors of the MorphIO API (like Morphology.points, Morphology.diameters).
My question is that is it possible to get the index of the starting points of each section directly from the API? I am trying to avoid loops in Python for performance issue, so it would be better to have such array directly constructed by the API.

Suggestions and discussions are appreciated. 

Cheers,
Weina

wizmer:
Hi,

Yes, I can add this as an attribute of the Morphology class. That will be a vector of int and each one will give you the offset in the point array. It will be available as `section_indices`. Sections will be in the depth first search order in a given neurite and the neurite order is the one of the file.

One thing to note is that Morphology.points only provides points belonging to neurites. Soma points can only be retrieved as `Morphology.soma.points`.

Does this sound good to you ?

WeinaJi:
Hello Benoit, 

The ```section_indices``` sounds very good. Thanks !

Yes, I see that soma points are not included in the Morphology.points. In SpatialIndex we can add soma and segments separately per gid. Are the ```Morphology.points``` always connected to the first element of the array ```Morphology.soma.points``` ?

wizmer:
No, the neurite point are not necessarily connected to the soma points. It may be, but it does not have to be.

PR #117, opened 2019-10-22
Add Morphology::sectionOffsets, the list of section offsets

wizmer:
In the immutable version, Morphology::points() and Morphology::diameters()
return a list of all points/diameters of the morphology. However it was no very
usable as the information about which point is in which section was not
available. See: https://github.com/BlueBrain/MorphIO/issues/116

This commits adds the method Morphology::sectionOffsets that returns the list
of offset to access the points of a given section.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ matrix:
       env:
         CXX: clang++
         CC: clang
+
     - os: osx
       osx_image: xcode10.1
       language: generic
@@ -19,13 +20,13 @@ matrix:
       env:
         CXX: clang++
         CC: clang
-     - os: osx
-       osx_image: xcode10.1
-       language: generic
-       python: 2.7
-       env:
-         CXX: g++
-         CC: gcc
+    - os: osx
+      osx_image: xcode10.1
+      language: generic
+      python: 2.7
+      env:
+        CXX: g++
+        CC: gcc
 
 cache:
   directories:
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -52,23 +52,37 @@ static void bind_immutable_module(py::module &m) {
         .def_property_readonly("points", [](morphio::Morphology* morpho){
                 return py::array(static_cast<py::ssize_t>(morpho->points().size()), morpho->points().data());
             },
-            "Returns a list with all points from all sections")
-        .def_property_readonly("section_offsets", [](morphio::Morphology* morpho){
-                return py::array(static_cast<py::ssize_t>(morpho->sectionOffsets().size()), morpho->sectionOffsets().data());
-            },
-            "Returns a list with offsets to access data of a specific section in the points and diameters arrays")
-        .def_property_readonly("diameters", [](morphio::Morphology* morpho){
-                auto diameters = morpho->diameters();
+            "Returns a list with all points from all sections (soma points are not included)\n"
+            "Note: points belonging to the n'th section are located at indices:\n"
+            "[Morphology.sectionOffsets(n), Morphology.sectionOffsets(n+1)[")
+        .def_property_readonly("diameters", [](const morphio::Morphology& morpho){
+                auto diameters = morpho.diameters();
                 return py::array(static_cast<py::ssize_t>(diameters.size()), diameters.data());
             },
-            "Returns a list with all diameters from all sections (soma points are not included)")
-        .def_property_readonly("perimeters", [](morphio::Morphology* obj){
-                auto data = obj->perimeters();
+            "Returns a list with all diameters from all sections (soma points are not included)\n"
+            "Note: diameters belonging to the n'th section are located at indices:\n"
+            "[Morphology.sectionOffsets(n), Morphology.sectionOffsets(n+1)[")
+        .def_property_readonly("perimeters", [](const morphio::Morphology& obj){
+                auto data = obj.perimeters();
                 return py::array(static_cast<py::ssize_t>(data.size()), data.data());
             },
-            "Returns a list with all perimeters from all sections")
-        .def_property_readonly("section_types", [](morphio::Morphology* obj){
-                auto data = obj->sectionTypes();
+            "Returns a list with all perimeters from all sections (soma points are not included)\n"
+            "Note: perimeters belonging to the n'th section are located at indices:\n"
+            "[Morphology.sectionOffsets(n), Morphology.sectionOffsets(n+1)[")
+        .def_property_readonly("section_offsets", [](const morphio::Morphology& morpho){
+                return py::array(static_cast<py::ssize_t>(morpho.sectionOffsets().size()),
+                                 morpho.sectionOffsets().data());
+            },
+            "Returns a list with offsets to access data of a specific section in the points\n"
+            "and diameters arrays.\n"
+            "\n"
+            "Example: accessing diameters of n'th section will be located in the DIAMETERS\n"
+            "array from DIAMETERS[sectionOffsets(n)] to DIAMETERS[sectionOffsets(n+1)-1]\n"
+            "\n"
+            "Note: for convenience, the last point of this array is the points array size\n"
+            "so that the above example works also for the last section.")
+        .def_property_readonly("section_types", [](const morphio::Morphology& morph){
+                auto data = morph.sectionTypes();
                 return py::array(static_cast<py::ssize_t>(data.size()), data.data());
             },
             "Returns a vector with the section type of every section")
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -50,20 +50,21 @@ static void bind_immutable_module(py::module &m) {
 
         // Property accessors
         .def_property_readonly("points", [](morphio::Morphology* morpho){
-                return py::array(static_cast<py::ssize_t>(morpho->points().size()), morpho->points().data());
+                const auto& data = morpho->points();
+                return py::array(static_cast<py::ssize_t>(data.size()), data.data());
             },
             "Returns a list with all points from all sections (soma points are not included)\n"
             "Note: points belonging to the n'th section are located at indices:\n"
             "[Morphology.sectionOffsets(n), Morphology.sectionOffsets(n+1)[")
         .def_property_readonly("diameters", [](const morphio::Morphology& morpho){
-                auto diameters = morpho.diameters();
-                return py::array(static_cast<py::ssize_t>(diameters.size()), diameters.data());
+                const auto& data = morpho.diameters();
+                return py::array(static_cast<py::ssize_t>(data.size()), data.data());
             },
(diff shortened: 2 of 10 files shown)


tests/test_1_swc.py
```diff
@@ -72,8 +73,8 @@ def test_read_simple():
     assert_equal(len(simple.root_sections[0].children), 2)
     assert_equal(simple.root_sections[0].children[0].id, 1)
     assert_equal(simple.root_sections[0].children[1].id, 2)
-    # assert_array_equal(simple.root_sections[0].children[0].points, [[0, 5, 0], [-5, 5, 0]])
-    # assert_array_equal(simple.root_sections[1].points, [[0, 0, 0], [0, -4, 0]])
+    assert_array_equal(simple.root_sections[0].children[0].points, [[0, 5, 0], [-5, 5, 0]])
```
wizmer: Unrelated to the current PR but I don't know why this was commented so I uncommented it.

include/morphio/morphology.h
```diff
@@ -82,6 +82,12 @@ class Morphology
      **/
     const Points& points() const;
 
+    /**
+     * Returns a list with offsets to access data of a specific section in the points
+     * and diameters arrays
+     **/
+    const std::vector<int> sectionOffsets() const;
```
WeinaJi: The output of section_offsets looks correct to me.
Just one small question here,  would it be better to use uint32_t for offset  so that it is consistent with the type of section_id?

wizmer: Good point, I've changed that.

.travis.yml
```diff
@@ -12,20 +12,21 @@ matrix:
...
-       env:
-         CXX: g++
-         CC: gcc
+    - os: osx
+      osx_image: xcode10.1
+      language: generic
+      python: 2.7
+      env:
+        CXX: g++
+        CC: gcc
```
mgeplf: what is providing gcc?  is it brew? I'm not sure this is needed in that case, as it should be covered by the linux plans, which run a lot quicker

binds/python/bind_immutable.cpp
```diff
@@ -53,6 +53,10 @@ static void bind_immutable_module(py::module &m) {
                 return py::array(static_cast<py::ssize_t>(morpho->points().size()), morpho->points().data());
             },
             "Returns a list with all points from all sections")
+        .def_property_readonly("section_offsets", [](morphio::Morphology* morpho){
+                return py::array(static_cast<py::ssize_t>(morpho->sectionOffsets().size()), morpho->sectionOffsets().data());
```
mgeplf: why the static cast to ssize_t when the type is uin32_t?

src/morphology.cpp
```diff
@@ -137,14 +137,26 @@ const Points& Morphology::points() const
 {
     return get<Property::Point>();
 }
+
+const std::vector<uint32_t> Morphology::sectionOffsets() const
+{
+    const std::vector<std::array<int, 2>>& indices_and_parents = get<Property::Section>();
```
mgeplf: auto?

src/morphology.cpp
```diff
@@ -137,14 +137,26 @@ const Points& Morphology::points() const
...
+    std::vector<uint32_t> indices(indices_and_parents.size());
+    std::transform(indices_and_parents.begin(), indices_and_parents.end(), indices.begin(),
+                   [](const std::array<int, 2>& pair) { return pair[0]; });
```
mgeplf: instead of `std::array<int, 2>`, shouldn't it be `Section::Type` - this applies everywhere.

binds/python/bind_immutable.cpp
```diff
@@ -53,6 +53,10 @@ static void bind_immutable_module(py::module &m) {
...
+            },
+            "Returns a list with offsets to access data of a specific section in the points and diameters arrays")
```
mgeplf: it would probably be helpful to describe this better - not everyone is used to the `h5` style encoding.  Maybe say that the `n`'th section is at `n` in the array, and that its points range from `points[section_offsets[n], section_offsets[n + 1] - 1]` , that the end has to be manually computed, and whether or not the parent point is included...

binds/python/bind_immutable.cpp
```diff
@@ -53,6 +53,10 @@ static void bind_immutable_module(py::module &m) {
                 return py::array(static_cast<py::ssize_t>(morpho->points().size()), morpho->points().data());
             },
             "Returns a list with all points from all sections")
+        .def_property_readonly("section_offsets", [](morphio::Morphology* morpho){
```
tristan0x: I am not sure, but I think pybind11 can accepts lambda taking `const MyClass&` in parameter. It would be safer here:
1. there is no need for pointer because the object is always there
1. the instance should not be modifiable

binds/python/bind_immutable.cpp
tristan0x: `Morphology::sectionOffsets` returns a new vector so:
* it should be called only once.
* because the vector is deallocated after the function and because if I remember correctly `py::array` does not copy the data buffer given in parameter, the returned `py::array` points to deleted memory. You may have to use a Python capsule to keep the vector alive as long as the Python object lives and pass a function pointer to Python so that it can destroyed the vector when Python GC destroyed the object. I think @ferdonline has a nice utility function to so that.

tests/test_4_immut.py
```diff
@@ -50,11 +50,15 @@ def test_iter():
...
                             cell.root_sections[0].children[0].iter(upstream)],
-                           [[[0.,  5.,  0.],
-                             [-5.,  5.,  0.]],
+                           [[[0., 5., 0.],
+                             [-5., 5., 0.]],
                             [[0., 0., 0.],
                              [0., 5., 0.]]])
 
+def test_section_offsets():
+    for _, cell in CELLS.items():
```
tristan0x:
```suggestion
    for cell in CELLS.values():
```
??

.travis.yml
wizmer: It does not seem to be brew: /Applications/Xcode-10.1.app/Contents/Developer/usr/bin/g++
Should I still remove it ?

mgeplf: I suppose it probably runs in parallel, so it won't reduce the build time.  Don't worry about it.

binds/python/bind_immutable.cpp
wizmer: If you don't you get:
>  /home/bcoste/workspace/io/binds/python/bind_immutable.cpp:53:55: error: conversion to ‘pybind11::ssize_t {aka long int}’ from ‘std::vector<std::array<float, 3> >::size_type {aka long unsigned int}’ may change the sign of the result [-Werror=sign-conversion]

src/morphology.cpp
wizmer: Until the day code editors will replace `auto` by the real type in the visual display, I find its more confusing than anything else for the reader.

binds/python/bind_immutable.cpp
wizmer: Cool, yes it works.

binds/python/bind_immutable.cpp
wizmer: Can I merge the current version and I'll use @ferdonline function in a seperate commit ?

WeinaJi:
Hello, 

May I ask the status of this PR? Is it ready for merging? Thanks!

Best, 
Weina

mgeplf:
I think @tristan0x's comments need to be addressed - pointing to freed'd memory for return values is a path to problems, so either that needs to be addressed first or done in this PR.

binds/python/bind_immutable.cpp
```diff
@@ -70,8 +70,7 @@ static void bind_immutable_module(py::module &m) {
             "Note: perimeters belonging to the n'th section are located at indices:\n"
             "[Morphology.sectionOffsets(n), Morphology.sectionOffsets(n+1)[")
         .def_property_readonly("section_offsets", [](const morphio::Morphology& morpho){
-                return py::array(static_cast<py::ssize_t>(morpho.sectionOffsets().size()),
-                                 morpho.sectionOffsets().data());
+                return as_pyarray(morpho.sectionOffsets());
```
mgeplf: sectionOffsets is declared:
`const std::vector<uint32_t> sectionOffsets() const;` (which is itself weird), but I'm surprised that `as_pyarray` deals with the `const` part without complaint - do we not have enough warnings on?

tristan0x: This is an incredible remark! Thank you very much @mgeplf 

The `const` return type of member function `sectionOffsets` is not a problem for `as_pyarray` as the constness it  part of the `sequence` **template** parameter type.

The real question here is "what is going on in the function body?"
```
Sequence* seq_ptr = new Sequence(std::move(seq));
```
In our case `Sequence` is `const std::vector<uint32>`. And because `std::move(seq)` is
a const rvalue reference, it is not possible to use the move constructor (that expects a non-const rvalue reference), and it falls back to ... the copy constructor!

Here is a snippet that highlights the case: https://godbolt.org/z/2qR9jb

I will remove the const return type of `sectionOffsets` in a minute.
Thanks again

mgeplf: > I will remove the const return type of sectionOffsets in a minute

Good call, I had the same thing in my version; I was using the version of libsonata's asArray, but it doesn't have the `static_cast<py::ssize_t>` in it, and I was trying to decide how annoyed I was by that...

All that to say you beat me to the PR, and I appreciate you doing it :)

binds/python/bindings_utils.cpp
```diff
@@ -65,3 +65,24 @@ static morphio::Points array_to_points(py::array_t<float> &buf){
...
+ * @brief "Casts" a Cpp sequence to a python array (no memory copies)
+ *  Python capsule handles void pointers to objects and makes sure
+ *      that they will remain alive.
+ *
+ *      https://github.com/pybind/pybind11/issues/1042#issuecomment-325941022
+ */
+template <typename Sequence>
+inline py::array_t<typename Sequence::value_type> as_pyarray(Sequence&& seq) {
+    // Move entire object to heap. Memory handled via Python capsule
+    Sequence* seq_ptr = new Sequence(std::move(seq));
```
mgeplf: can you add `#include <utility>  //for std::move`, I like files to include their deps when possible

tristan0x: sure, thanks

src/morphology.cpp
```diff
@@ -79,27 +79,27 @@ Morphology::~Morphology()
...
 
-const Mitochondria Morphology::mitochondria() const
+Mitochondria Morphology::mitochondria() const
 {
-    return Mitochondria(_properties);
+    return {_properties};
 }
 
-const std::vector<Property::Annotation> Morphology::annotations() const
+const std::vector<Property::Annotation>& Morphology::annotations() const
```
mgeplf: sharp catch

Issue #85, opened 2019-07-04
Access to connectivity information without generating the sections (immutable)


eleftherioszisis:
My understanding for the immutable Morphology is that the section connectivity is stored in a table.

Currently, one can access the morphology points and properties, such as diameters, but there is not way to access the section connectivity from the morphology level.

The section connectivity could be returned in the form of an edge list e.g.
```python
from morphio import Morphology
morphology = Morphology(path)
section_connectivity = morphology.section_connectivity
```
which would return an array like the following:
```
[[0, 1], [1, 2], [1, 3]]
```

Expanding on this, given that the section has an associated id, then we could retrieve the specific section by id, while having the section node adjacency.

Returning also the point connectivity can be useful for graph analysis. For example:
```python
from morphio import Morphology
morphology = Morphology(path)
points = morphology.points
point_connectivity = morphology.point_connectivity
```
This would allow to directly convert the morphology point connectivity into an adjacency matrix of the vertices / points.

arnaudon:
Yes, that would be really nice to have. I'm interested in this from a 'math' perspective of understanding the morphological properties/features of neurons.

I think that a good option could be the following:
return a .csv file (or something like that), where each line has the following:
- id_a: id of the point nearest to soma
- id_b: id of the other point forming the segment
- lengths: lengths of the segment
- radius: mean (?) radius or radius_a/radius_b...
- other meaningful attributes of segments... 

In addition, if there are any interesting point attributes, we could return another .csv, with a list of points, there id corresponding to the ids in the segment list, and their attributes. 

Thanks!

PR #118, opened 2019-10-22
Add Morphology::connectivity

wizmer:
which returns the connectivity between sections

### DIFF
--- a/src/mut/morphology.cpp
+++ b/src/mut/morphology.cpp
@@ -318,7 +318,7 @@ void Morphology::applyModifiers(unsigned int modifierFlags) {
 std::map<int, std::vector<unsigned int>> Morphology::connectivity()
 {
     std::map<int, std::vector<unsigned int>> connectivity;
-    for(auto kv: _children)
+    for(const auto& kv: _children)
     {
         auto node_id = static_cast<int>(kv.first);
         connectivity[node_id] = std::vector<unsigned int>();
--- a/src/mut/morphology.cpp
+++ b/src/mut/morphology.cpp
@@ -315,20 +315,22 @@ void Morphology::applyModifiers(unsigned int modifierFlags) {
         modifiers::nrn_order(*this);
 }
 
-std::unordered_map<int, std::vector<unsigned int>> Morphology::connectivity()
-{
+std::unordered_map<int, std::vector<unsigned int>> Morphology::connectivity() {
     std::unordered_map<int, std::vector<unsigned int>> connectivity;
 
     const auto& roots = rootSections();
     connectivity[-1].reserve(roots.size());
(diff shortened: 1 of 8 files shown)


wizmer:
Cf https://github.com/BlueBrain/MorphIO/issues/85

src/mut/morphology.cpp
```diff
@@ -388,6 +388,30 @@ void Morphology::applyModifiers(unsigned int modifierFlags)
         modifiers::nrn_order(*this);
 }
 
+std::map<int, std::vector<unsigned int>> Morphology::connectivity()
+{
+    std::map<int, std::vector<unsigned int>> connectivity;
+    for(auto kv: _children)
+    {
+        int node_id = static_cast<int>(kv.first);
+        connectivity[node_id] = std::vector<unsigned int>();
```
tristan0x: `map::operator[]` adds the element if not found, so instruction L397 is not required. Besides:
1. the loop over `kv.second` will perform the same access to `connectivity[node_id]` over and over.
1. the assigned vector memory space should be reserved before _push_backing_.
1. the `section` loop variable should be a const reference.
1. use the STL :)

Something like:
```cpp
auto& vec = connectivity[node_id];
vec.reserve(kv.second.size());
std::transform(kv.second.begin(), kv.second.end(), std::back_inserter(vec), [](const auto& section) { return section->id; });
```
Note: the `auto` in the lambda is C++14 only.

tests/test_4_immut.py
```diff
@@ -56,6 +56,11 @@ def test_iter():
                              [0., 5., 0.]]])
 
 
+def test_connectivity():
+    for _, cell in CELLS.items():
```
tristan0x:
```suggestion
    for cell in CELLS.values():
```
??

tests/test_5_mut.py
```diff
@@ -40,6 +41,17 @@ def test_point_level():
...
 
 
+def test_connectivity():
+    cells = OrderedDict({
+        'asc': Morphology(os.path.join(_path, "simple.asc")),
+        'swc': Morphology(os.path.join(_path, "simple.swc")),
+        'h5': Morphology(os.path.join(_path, "h5/v1/simple.h5")),
+    })
+
+    for _, cell in cells.items():
```
tristan0x:
```suggestion
    for cell in cells.values():
```
??

src/mut/morphology.cpp
wizmer: Thanks. About 1), it's probably optimized anyway by the compiler no ?

tristan0x: Yes, in theory, the compilation caches the access because it "sees" that nothing modifies `connectivity` inside it. But I prefer not to bet on that.

arnaudon:
Would it be possible to get connectivity between points? (especially for vasvulature)

PR #119, opened 2019-11-12
Add test scaffold

mgeplf:


cpp_test.sh
```diff
@@ -0,0 +1,14 @@
...
+# This file runs the C++ tests, as well as compiling the code with warnings on
+# so that errors should be caught quicker
+
+set -euxo pipefail
+
+rm -rf build;
+mkdir build 
+pushd build 
+cmake .. -DMorphIO_CXX_WARNINGS=ON
+make -j2
```
wizmer: `make -j`, no ? YOLO ?

tests/main.cpp
```diff
@@ -0,0 +1,2 @@
+#define CATCH_CONFIG_MAIN
```
wizmer: Can you add a docstring about the purpose of this file. In particular what does it bring wrt test_morphology.cpp

cpp_test.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# This file runs the C++ tests, as well as compiling the code with warnings on
```
wizmer: That's really cool

wizmer:
I just discovered that we were actually testing the macox builds with python 2, even when specifying a mac os which defaults to python 3. You need to explicitly call python 3 when creating the virtual env.
https://github.com/BlueBrain/MorphIO/pull/130/commits/6189d7758512c04bf90887b1be633de9283cb543

PR #125, opened 2019-12-05
Don't write comment at end of SWC file

mgeplf:
* so that NEURON doesn't raises an error message (I have a PR to fix the blank line issue in their SWC parser
* move local functions to anon ns
* remove unused iostream header
* no need to call open/close on ofstreams


wizmer:
Out of curiosity, can you point me at the NEURON PR ?

mgeplf:
https://github.com/neuronsimulator/nrn/pull/348

Issue #127, opened 2019-12-11
Writing an h5 results in seg-fault if annotations are present on morphology


lidakanari:
Example (swc works but annotations are not written to file, h5 throws seg-fault error)
```
import morphio
m = morphio.mut.Morphology()
m.annotations.append('{"Apical point": (0, 1, 0)}')

m.write('Test annot.swc')
Warning: writing file without a soma

m.write('Test annot.h5')
Warning: writing file without a soma
Segmentation fault
```

wizmer:
Annotations are not designed to be writable, it's a read-only object.
They just contain details about what happened when the files was opened (like single children, duplicates...).
And the segfault comes from the fact you are trying to write an empty morphology.

tristan0x:
> And the segfault comes from the fact you are trying to write an empty morphology.

IMHO the library should not punish the user with a segfault.

wizmer:
> IMHO the library should not punish the user with a segfault.
Yep. This has been fixed.

PR #128, opened 2019-12-16
Skip writing when calling "write" on an empty morphology

wizmer:
Issue was mentioned in https://github.com/BlueBrain/MorphIO/issues/127


include/morphio/enums.h
```diff
@@ -33,6 +33,7 @@ enum Warning
     UNDEFINED,
     MITOCHONDRIA_WRITE_NOT_SUPPORTED,
     WRITE_NO_SOMA,
+    WRITE_EMPTY_MORPHOLOGY,
```
tristan0x: I would append the new `WRITE_EMPTY_MORPHOLOGY` enum value at the end to prevent shifting all existing identifiers.

src/errorMessages.cpp
```diff
@@ -249,6 +249,11 @@ std::string ErrorMessages::WARNING_WRITE_NO_SOMA() const
     return "Warning: writing file without a soma";
 }
 
+std::string ErrorMessages::WARNING_WRITE_EMPTY_MORPHOLOGY() const
+{
+    return "Warning: attempting to write an empty morphology. Refusing to write.";
```
tristan0x: _Refusing_  ? I would not be so rude with the user :)
What about something like: "Warning: aborting an attempt to write an empty morphology."  ?

tristan0x: Some warning messages start with "Warning: ", others don't. IMO they should be consistent.

mgeplf:
FYI, I don't think you need to have:
```
* fixup! Prefer std::ostringstream to std::string concatenation
* fixup! Fix regression in number formatting of error message
* fixup! git-clang-format bd18496~1
``
in the commit messages: https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

PR #130, opened 2019-12-18
Accept any string-like object as filename argument

wizmer:
That's especially helpful with to load Morphologies from `pathlib.Path` objects.


binds/python/bind_immutable.cpp
```diff
@@ -20,10 +20,16 @@ static void bind_immutable_module(py::module &m) {
...
 
     py::class_<morphio::Morphology>(m, "Morphology")
-        .def(py::init<const morphio::URI&, unsigned int>(),
+        .def(py::init<const std::string&, unsigned int>(),
              "filename"_a, "options"_a=morphio::enums::Option::NO_MODIFIER)
         .def(py::init<morphio::mut::Morphology&>())
-
+        .def(py::init([](py::object arg, unsigned int options) {
+                          return std::unique_ptr<morphio::Morphology>(
+                              new morphio::Morphology(py::str(arg), options)
```
mgeplf: what happens when `py::str()` doesn't work?

include/morphio/types.h
```diff
@@ -14,7 +14,6 @@
 
 /** @namespace morphio Blue Brain File IO classes */
 namespace morphio {
-using URI = std::string;
```
mgeplf: woohoo

binds/python/bind_immutable.cpp
wizmer: I don't manage to find an example of object for which it does not work.
```
class A:
    pass
morphio.Morphology(A())
```
The error is `RawDataError: File: <__main__.A object at 0x7fc49a349470> does not exist.`
Because `str(A())` still returns something even though `str` is not explicitly defined.
But I think the error message is still helpful.

mgeplf: yeah, on further thought, `str()` is basically defined for everything

tests/test_10_vasculature.py
```diff
@@ -67,3 +68,8 @@ def test_iterators_vasculature():
     for sec in morphology.iter():
         all_sections.remove(sec.id)
     assert_equal(len(all_sections), 0)
+
+
+def test_from_pathlib():
+    vasc = vasculature.Vasculature(Path(_path, "h5/vasculature1.h5"))
```
tristan0x: neat!

mgeplf:
LGTM

PR #131, opened 2019-12-20
Check clang format

mgeplf:
* github action checks on push that changes are formatted correctly
* applied clang format to full all code

### DIFF
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -1,3 +1,5 @@
+#include "bind_immutable.h"
+
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 #include <pybind11/iostream.h>  // py::add_ostream_redirect
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -1,20 +1,18 @@
 #include <pybind11/pybind11.h>
 
-#include <pybind11/iostream.h>
-#include <pybind11/numpy.h>
-#include <pybind11/operators.h>
-#include <pybind11/stl.h>
-#include <pybind11/stl_bind.h>
+#include <pybind11/iostream.h>  // py::add_ostream_redirect
 
 #include <morphio/enums.h>
 #include <morphio/mut/morphology.h>
+#include <morphio/soma.h>
 #include <morphio/types.h>
 
 #include "bind_enums.h"
+#include "bindings_utils.h"
 
 namespace py = pybind11;
 
-static void bind_immutable_module(py::module& m) {
+void bind_immutable_module(py::module& m) {
     using namespace py::literals;
 
     // http://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html?highlight=iostream#capturing-standard-output-from-ostream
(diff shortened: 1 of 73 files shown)


mgeplf:
I haven't changed any C++ code, and the build is failing...

binds/python/bind_immutable.cpp
```diff
@@ -1,7 +1,8 @@
+#include <pybind11/pybind11.h>
```
tristan0x: I am sure a good Samaritan will break the build by revert this.
Let me try to disable the formatting of the line before merging this.
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code

tristan0x: It is only possible to disable formatting within a range. Nothing elegant for a single line.

```diff
diff --git a/binds/python/bind_immutable.cpp b/binds/python/bind_immutable.cpp
index b9209a3..f12909d 100644
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -1,5 +1,6 @@
+// clang-format off
 #include <pybind11/pybind11.h>
-
+// clang-format on
 #include <pybind11/iostream.h>
 #include <pybind11/numpy.h>
 #include <pybind11/operators.h>
```

Would this work for you @mgeplf ?

If not, IMO we should at least document the rationale:
```diff
diff --git a/binds/python/bind_immutable.cpp b/binds/python/bind_immutable.cpp
index b9209a3..9366ca3 100644
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -1,3 +1,4 @@
+// top-level pybind11 header must be the included beforehand
 #include <pybind11/pybind11.h>
 
 #include <pybind11/iostream.h>
```

mgeplf: yeah, good call; I think I also need to clean up how this works b/c of the includes of `.cpp` and such files, but I'll do that in another commit

mgeplf: I should note that I *think* that clang format only sorts within blocks; so if there is a line break, it shouldn't be reverted.  I'll leave a comment though

tristan0x: > I should note that I _think_ that clang format only sorts within blocks; so if there is a line break, it shouldn't be reverted.

Absolutely.

> I'll leave a comment though

:+1:

ci/check_clang_format.sh
```diff
@@ -0,0 +1,26 @@
...
+    "$VENV/bin/pip" install clang-format=="$CLANG_FORMAT_VERSION"
+fi
+
+set +u  # ignore errors in virtualenv's activate
+source "$VENV/bin/activate"
+set -u
+
+changes=$(git-clang-format 'HEAD~1')
+if [[ $(echo "$changes" | grep -n1 'changed files') ]]; then
+    echo "The following files require changes to pass the current clang-format"
```
wizmer: Can you specify what action the user should take when this happen.
I guess it is to run the thing locally and commit ?

mgeplf: pretty much

Issue #126, opened 2019-12-09
Writing morphology with perimeters as ascii does not throw an error


eleftherioszisis:
ascii format does not support perimeters (only h5). In that case shouldn't morphio raise an error if someone (like me) mistakenly wrote the file in a different format? 

Example:
``` python
import morphio
m = morphio.Morphology('test.asc').as_mutable()

for section in m.iter():
    section.perimeters = section.diameters

m.write('test_perimeters.asc')
```
It silently drops perimeters and writes the file.

PR #135, opened 2019-12-23
Raise when attempting to write cell with perimeter data to SWC or ASC

wizmer:
and add missing binding for morphio::WriterError exception

Issue #126


tests/requirement_tests.txt
```diff
@@ -1,3 +1,4 @@
 mock>=2.0.0
 nose>=1.3.7
 numpy>=1.14.2
+h5py>=2.9.0
```
tristan0x: I would keep the file sorted.

tests/test_6_writers.py
```diff
@@ -69,6 +72,10 @@ def test_write_basic():
         assert_array_equal(ImmutMorphology(os.path.join(tmp_folder, "test_write.swc")).points, expected)
         assert_array_equal(ImmutMorphology(os.path.join(tmp_folder, "test_write.h5")).points, expected)
 
+        with h5py.File(h5_out) as h5_file:
+            ok_(not '/perimeters' in h5_file.keys())
```
tristan0x: pep8 guideline
```suggestion
            ok_('/perimeters' not in h5_file.keys())
```

Besides, `.keys()` should not be required.
```suggestion
            ok_('/perimeters' not in h5_file)
```

src/mut/writers.cpp
```diff
@@ -75,6 +82,9 @@ void swc(const Morphology& morphology, const std::string& filename)
         return;
     }
 
+    if (hasPerimeterData(morphology))
```
tristan0x: IMO it is more readable and less error-prone to put braces even around statements with one instruction.
Besides, there is a [clang-tidy check for that](https://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html) I recommend.

src/mut/writers.cpp
```diff
@@ -171,6 +181,9 @@ void asc(const Morphology& morphology, const std::string& filename)
...
```
tristan0x: same remark about braces.

src/mut/writers.cpp
```diff
@@ -319,7 +330,7 @@ void h5(const Morphology& morpho, const std::string& filename)
         // If the morphology has some perimeter data, we need to fill some
         // perimeter dummy value in the soma range of the data structure to keep
         // the length matching
-        if (hasPerimeterData)
+        if (hasPerimeterData_)
```
tristan0x: same remark about braces.

src/mut/writers.cpp
```diff
@@ -368,7 +379,7 @@ void h5(const Morphology& morpho, const std::string& filename)
     write_attribute(h5_file, "comment",
         std::vector<std::string>{version_string()});
 
-    if (hasPerimeterData)
+    if (hasPerimeterData_)
```
tristan0x: same remark about braces.

src/mut/writers.cpp
```diff
@@ -29,6 +29,13 @@ struct base_type<std::vector<T>> : base_type<T>
 {
 };
 
+bool hasPerimeterData(const morphio::mut::Morphology& morpho)
+{
+    return !morpho.rootSections().empty()
```
alkino: `return !morpho.rootSections().empty() &&!morpho.rootSections()[0]->perimeters().empty() ;`

seems me more common

src/mut/writers.cpp
```diff
@@ -29,6 +29,13 @@ struct base_type<std::vector<T>> : base_type<T>
 {
 };
 
+bool hasPerimeterData(const morphio::mut::Morphology& morpho)
```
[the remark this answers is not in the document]
wizmer: I'm not a big fan of const on primitive types. But I tried and gcc gave me a warning:
`error: type qualifiers ignored on function return type [-Werror=ignored-qualifiers]`

src/errorMessages.cpp
```diff
@@ -241,6 +241,11 @@ std::string ErrorMessages::ERROR_VECTOR_LENGTH_MISMATCH(const std::string& vec1,
     return msg;
 }
 
+std::string ErrorMessages::ERROR_PERIMETER_DATA_NOT_WRITABLE()
```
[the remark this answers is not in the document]
wizmer: I would have to change them all then. I will do in another commit.

src/mut/writers.cpp
wizmer: Thanks.

src/mut/writers.cpp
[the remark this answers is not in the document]
alkino: Hum

I talk of
bool hasPerimeterData(const morphio::mut::Morphology& morpho) const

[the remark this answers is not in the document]
wizmer: It's not a method in fact. Just a function.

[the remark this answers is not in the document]
alkino: oups sorry for that

PR #140, opened 2020-01-09
Fix README TOC

alkino:


alkino:
Maybe people knowing this library should add example of the Read/Write API in README. Right now it only shows includes needed.

wizmer:
Thanks for the PR !
What do you mean with the examples ? You mean, expand the one at https://github.com/BlueBrain/MorphIO/tree/fix_readme#creating-morphologies-with-the-mutable-api ?

PR #141, opened 2020-01-10
Do not use pybind11::make_iterator default value for template parameter ValueType 

wizmer:
As this used to cause a bug in python2 when calling iterators with zip. 
See the added unit test to see what was failing.


tristan0x:
AFAIU, this patch only fixes a use-case with Python 2.

If I understand correctly [BBP support policy](https://bbpteam.epfl.ch/project/spaces/display/BSD/Software+Deployment+Strategy+and+Support) 
, versions of Python < 3.6 are only supported at best-effort.

So I do not understand why we should introduce a performance penalty to all other use-cases (both C++ and Python 3 API) for an unsupported use-case.

If you really want to keep Python 2 support, what about using a preprocessor macro to control the `value_type` type of the iterator classes, for instance:

```c++
template <typename SectionT, typename MorphologyT>
class breadth_iterator_t
{
  public:
#ifdef MORPHIO_SECTION_ITERATOR_USE_LVALUE
    using value_type = SectionT;
#else
   using value_type = const SectionT&;
#endif // MORPHIO_SECTION_ITERATOR_USE_LVALUE

    // [...]
  
    inline value_type operator*() const;

    // [...]
};
```

In this case, the macro would be defined only when bindings are requested and Python is version 2.

mgeplf:
> AFAIU, this patch only fixes a use-case with Python 2.

It's showing up in the tests as failing for py2, but I don't understand why it isn't happening for py3 - do you have any idea why there is a difference? I assume it's how pybind11 is handling it, but I don't see why it would be different for py2 vs. py3.

I probably wouldn't name a reference as `value_type = const SectionT&;`, though.

tristan0x:
> I probably wouldn't name a reference as `value_type = const SectionT&;`, though.

I thought it was defined in the standard but I was wrong. According to [iterator](http://www.cplusplus.com/reference/iterator/iterator/) definition, it should look more like below:

```c++
template <typename SectionT, typename MorphologyT>
class breadth_iterator_t
{
  public:
    using value_type = const SectionT;
    using reference = value_type&;
    using pointer = value_type*;

#ifdef MORPHIO_SECTION_ITERATOR_USE_LVALUE
    inline value_type operator*() const;
#else
    inline reference operator*() const;
#endif // MORPHIO_SECTION_ITERATOR_USE_LVALUE
    // [...]
};
```

mgeplf:
> AFAIU, this patch only fixes a use-case with Python 2.

So I looked into this a bit more; and it's not just py2.  Internally, the iterators are keeping `deque`'s of copies of of `SectionT`: thus, when the iterator is dtor'd, that deque no longer exists, and thus the reference goes stale - I believe the others are working just because the vector memory hasn't been cleared.

wizmer:
Sorry my Emacs at home is going FUUUUUUULL retard on the indent levels.

wizmer:
I changed the default value of the template argument ValueType.
https://github.com/pybind/pybind11/blob/b4e5d582cb656a590b256bcf4a8ffa7c8ce9ba19/include/pybind11/pybind11.h#L1678
The default value is the returned type of the C++ iterator so that means when using a `const ref` the python object would still be a `const ref`. By changing it, it forces the python object to be a copy.

wizmer:
From the quick chat I had with @mgeplf this morning, he seems to think that the problem is even on the C++ side. According to him, the object returned by the iterator should never be a reference as it ties its lifetime to the one of the iterator. (@mgeplf , I hope I did not misinterpret your thought)
@tristan0x Do you think the iterator should always return a copy ?

mgeplf:
I wasn't clear enough in my previous comment, but the point is that internally:

> Internally, the iterators are keeping deque's of copies of of SectionT

So, when the iterator is dtor'd, or changed, any `const&` that come from iterator deref are invalid.  ie:
```
auto it = m.depth_begin();
const auto& sec0 = *it;  // ref to it::deque_.front();
++it; // calls it::deque_.pop_front(), so accessing sec0 is UB, no?
```

wizmer:
Ahhhhh, got it, I'm sorry it took me so long. So the return value is not only coupled to the lifetime of the iterator but to its state as well. That's плохой. 
So the 2 options are return a copy or a smart pointer ?

mgeplf:
I think a copy makes sense (a Section isn't really that big of an object), but I'd like to have @tristan0x comment.

tristan0x:
Thank you for the time spent investigating this. I understand now why `iterator::value_type` and its iterator must have an independent lifetime because they are attached to independent Python objects.

But for the consumers of MorphIO C++ API, IMO it is a pity to introduce a performance penalty by introducing a useless copy for 99% of the cases where the `iterator::value_type` won't live outside of its iterator's scope. But I am also worried about the remaining 1% where the user has the responsibility to make a copy. Would a Doxygen be enough?

In conclusion, my suggestion would be to:
1. keep the C++ API as it is, but add a comment warning users about that the lifetime of the const reference returned by `operator*` is bound to the iterator, not the associated container.
2. create an iterator proxy class whose purpose is to wrap any iterator and replace the `operator*` to return a copy.
3. use this proxy in the Python bindings to wrap arguments given to ` py::make_iterator`.

Here is a stupid example of proxy iterator https://godbolt.org/z/RZzCX_

mgeplf:
> But for the consumers of MorphIO C++ API, IMO it is a pity to introduce a performance penalty by introducing a useless copy for 99% of the cases where the iterator::value_type won't live outside of its iterator's scope

It's not just the scope; it's also advancing the iterator.

I don't think a comment will cut it as a way to warn people away from this - the fact that `py::bind` different between py2 & py3 happens to trigger it shows that it's a subtle thing to handle.

Finally, how expensive is this copy?  I'd like to see hard numbers before saying that it makes that much of a difference.

tristan0x:
> > But for the consumers of MorphIO C++ API, IMO it is a pity to introduce a performance penalty by introducing a useless copy for 99% of the cases where the iterator::value_type won't live outside of its iterator's scope
> 
> It's not just the scope; it's also advancing the iterator.
> 
> I don't think a comment will cut it as a way to warn people away from this - the fact that `py::bind` different between py2 & py3 happens to trigger it shows that it's a subtle thing to handle.
> 
> Finally, how expensive is this copy? I'd like to see hard numbers before saying that it makes that much of a difference.

Less is better. In some cases, using a reference is optimal, so I am not in favor to penalize them for the other cases where a copy must be used.

There is a cost for sure, but definitely insignificant compared to the I/O made by this library.

wizmer:
I don't know if my test (with -O3) sucks or not but:
```
TEST_CASE("time", "[morphology]") {
    const morphio::Morphology m("data/h5/v1/Neuron.h5");
    std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
    std::vector<unsigned int> vec;
    const auto &root = m.section(0);
    for(int i = 0; i<1000; ++i)
        for(auto it = root.depth_begin(); it != root.depth_end(); ++it)
            vec.push_back((*it).id());
    std::cout << "vec.size(): " << vec.size() << std::endl;

    std::chrono::steady_clock::time_point end = std::chrono::steady_clock::now();

    std::cout << "Time difference = " <<
        std::chrono::duration_cast<std::chrono::microseconds>(end - begin).count() << "[µs]" <<
        std::endl;
}
```
gives exactly the same times when returning a copy of a const ref.
I would be more inclined to go towards the copy version then.

PR #145, opened 2020-01-15
mut.Morphology.write accept anything that looks like a string

wizmer:


tests/test_6_writers.py
```diff
@@ -32,6 +33,8 @@ def test_write_soma_basic():
 
     with setup_tempdir('test_write_soma_basic') as tmp_folder:
         morpho.write(os.path.join(tmp_folder, "test_write.swc"))
+        morpho.write(Path(tmp_folder, "test_write_pathlib.swc"))
```
wizmer: Sorry, I became a pathlib junkie.

mgeplf:
Are there any other Path like places that this needs to be done to? Would be nice to get them all at once.

tests/test_6_writers.py
tristan0x: Careful to the ad-`dict`-ion

wizmer:
I think we are good now.

binds/python/bind_mutable.cpp
```diff
@@ -115,7 +115,9 @@ void bind_mutable_module(py::module& m) {
         .def_property_readonly("version", &morphio::mut::Morphology::version, "Returns the version")
 
         .def("write",
-             &morphio::mut::Morphology::write,
+             [](morphio::mut::Morphology* morph, py::object arg) {
```
tristan0x: I use pointers only when it is necessary to manipulate the object's address rather than the object itself. But this could be done in a separate PR to apply this to all bindings.

So more something like this:
```c++
             [](morphio::mut::Morphology& morph, const py::object& arg) {
```

wizmer:
I'm merging and will apply @tristan0x's comment in another PR

PR #149, opened 2020-01-29
Rely on pathlib2 package instead of standard pathlib module

wizmer:
By using pathlib2 instead of pathlib

### DIFF
--- a/ci/python_test.sh
+++ b/ci/python_test.sh
@@ -5,7 +5,7 @@ set -euxo pipefail
 VENV=env
 
 rm -rf "$VENV"
-virtualenv -p $(which python) "$VENV"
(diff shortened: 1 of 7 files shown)


ci/python_test.sh
```diff
@@ -5,7 +5,7 @@ set -euxo pipefail
 VENV=env
 
 rm -rf "$VENV"
-python -mvenv "$VENV"
+virtualenv -p $(which python) "$VENV"
```
mgeplf: newer python installs don't necessarily have virtualenv, but py3 is guaranteed to have -mvenv

wizmer: I have reverted my changes. I'll create a separate issue if we want to add python2 CI.

PR #150, opened 2020-01-29
Add CONTRIBUTING.md document

tristan0x:


wizmer:
That's awesome. Thanks a lot. Given that most of the CONTRIBUTING page is common to all our projects, I wonder if it could be shared among the BlueBrain repos.

Issue #148, opened 2020-01-28
When creating a morphio.mut.Morphology, soma.type is set to undefined and is unchangable.


wizmer:

wizmer:
When opening a morphology from a file, there is an heuristic that determines the soma type based on the number of points.
https://github.com/BlueBrain/MorphIO/blob/master/src/morphology.cpp#L182

Unlike most of the properties of the morphio.mut.Morphology object, the `soma.type` is readonly as it is controlled by the heuristic described above.

The problem is that when people are creating a new `morphio.mut.Morphology` from scratch, the heuristic can not be applied (as the neuron is created empty) and so the `soma.type` is set to `undefined`. That can be a problem for high-level softwares that want to use the newly created morphology without having to pass by the intermediate step of writing it to disk.

One *proposal* (feel free to discuss it), would be that when `soma.type()` (https://github.com/BlueBrain/MorphIO/blob/master/include/morphio/mut/soma.h#L77) is queried and it is `undefined` the heuristic is called, the soma type is updated and then returned.

Since the heuristic depend on the number of soma points, maybe the soma type should also be updated every time the points array is modified.

lidakanari:
I think it is reasonable to update the soma type if the points array is modified. Imagine that someone would like to correct the soma from one point to a contour. It is reasonable in this case to check the new type and update it accordingly.

tristan0x:
> One _proposal_ (feel free to discuss it), would be that when `soma.type()` (https://github.com/BlueBrain/MorphIO/blob/master/include/morphio/mut/soma.h#L77) is queried and it is `undefined` the heuristic is called, the soma type is updated and then returned.

This `Soma::_somaType` member variable is a pain because it must be updated whenever the number of points change. What about removing it and systematically call the heuristic you mentioned above in `Soma::type()` member function?

Besides I would use a `switch` statement in function `getSomaType` and make it `noexcept` instead of try/catch-ing the `std::map` (which is creating at every call).

arnaudon:
If I understood this well, when a neuron is loaded fro a file, the soma type is detected automatically. Could you implement a `neuron.update_soma_type()` function which would trigger this type detection, and assign the correct type? I guess this function could also be called whenever the soma points are modified.

wizmer:
Yep, I think I agree with @tristan0x that we could simply rely on the heuristic all the time and get ride of `Soma::_somaType`

eleftherioszisis:
+1 for the on the fly type deduction

[...]

PR #156, opened 2020-01-31
attempt on issue #148 on soma type

arnaudon:
Only a partial update, the test fail on the .h5 format.

### DIFF
--- a/include/morphio/mut/soma.h
+++ b/include/morphio/mut/soma.h
@@ -7,8 +7,7 @@ namespace mut {
 class Soma
 {
   public:
-    Soma()
-        : _somaType(SOMA_UNDEFINED) {}
+    Soma() {}
 
     Soma(const Property::PointLevel& pointProperties);
     Soma(const Soma& soma);
@@ -54,7 +53,6 @@ class Soma
 
   private:
     friend class Morphology;
-    SomaType _somaType;
     Property::PointLevel _pointProperties;
 };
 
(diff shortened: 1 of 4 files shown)


include/morphio/mut/soma.h
```diff
@@ -7,8 +7,7 @@ namespace mut {
 class Soma
 {
   public:
-    Soma()
-        : _somaType(SOMA_UNDEFINED) {}
+    Soma() {}
```
tristan0x:
```suggestion
    Soma() = default;
```

include/morphio/mut/soma.h
```diff
@@ -75,7 +73,15 @@ const std::vector<float>& Soma::diameters() const noexcept {
 }
 
 inline SomaType Soma::type() const noexcept {
-    return _somaType;
+    // this function seems not complete, see /src/readers/morphologySWC.cpp, line 217 for a more complete one
```
tristan0x: I would rather use a `switch` instead of allocating a `std::map` all the time and rely on an expensive exception raised by `std::map::at`.

Something like:
```c++
switch(points().size()) {
  case 0:
  case 2:
    return SOMA_UNDEFINED;
  case 1:
    return SOMA_SINGLE_POINT;
  default:
    return SOMA_SIMPLE_CONTOUR;
}
```

Besides the member function can then be marked with `noexcept` specifier.

cattabiani: what do you think about creating a static std::array<SomaType> v = {SOMA_UNDEFINED, SOMA_SINGLE_POINT, SOMA_UNDEFINED}; and pick from that if nPoints are inside the range? There is only one size check; it should be faster

arnaudon:
Guys, I'm not sure this is a PR worse working on. it does not pass the tests, and @wizmer told me it is a difficult issue to solve. He implemented a hack elsewhere to get around this bug for the time being. Maybe work on the other PR instead of this one -- I was bluffing on my c++ knowledge to try to fix this, but kind of failed... (but my bug was fixed :) )

PR #157, opened 2020-01-31
Implement Tristan's proposal Fix: #148

cattabiani:
The soma type was stored in a variable in Soma that was readonly.
When the Neuron was created from scratch the type was set UNDEFINED
(since the neuron wa empty) and there was no way of changing it
afterwards. Now we rely entirely on the heuristic function
morphio::getSomaType() and removed the variable _somaType alltogether.

getSomaType() was also simplified as Tristan suggested: the map
creation&search has been substituted with a switch simplifiing the
code and adding the noexcept operator.

The change was implemented in mut::Soma as well as in Soma. Since
getSomaType() is O(1) it should not degrade performances.

- Remove somaType_ variable from Soma and mut::Soma.
- Simplify morphio::getSomaType adding noexcept for extra security.

### DIFF
--- a/include/morphio/morphology.h
+++ b/include/morphio/morphology.h
@@ -12,6 +12,8 @@ enum SomaClasses { SOMA_CONTOUR, SOMA_CYLINDER };
 using breadth_iterator = breadth_iterator_t<Section, Morphology>;
 using depth_iterator = depth_iterator_t<Section, Morphology>;
 
+SomaType getSomaType(long unsigned int nSomaPoints) noexcept;
+
 /** Read access a Morphology file.
  *
  * Following RAII, this class is ready to use after the creation and will ensure
--- a/include/morphio/mut/soma.h
+++ b/include/morphio/mut/soma.h
@@ -7,8 +7,7 @@ namespace mut {
 class Soma
 {
   public:
-    Soma()
-        : _somaType(SOMA_UNDEFINED) {}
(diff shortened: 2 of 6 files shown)


wizmer:
Hi @cattabiani !

Thanks for the PR ! It is really great to see new people are interested by MorphIO.
I have 2 comments regarding your PR.
1) In the immutable case it's still O(1), but it is performing unnecessary re-computation as the immutability makes us certain that the soma type is not going to change. I would prefer to keep using the old way for the immutable version (also because the immutable API is really the one for which we want maximum performance) and only use the new heuristic on the mutable API.

2) I apologize for not noticing when I wrote the ticket but the heuristic actually depends on the file format.
In case we open the file from SWC, the heuristic is slightly different: https://github.com/BlueBrain/MorphIO/blob/master/src/readers/morphologySWC.cpp#L217
In short, if a soma has 3 points or more, if it comes from a H5/ASC file, it is a SOMA_CONTOUR
From a SWC, it is a SOMA_CYLINDERS or SOMA_NEUROMORPHO_THREE_POINT_CYLINDERS  https://github.com/BlueBrain/MorphIO/blob/master/doc/specification.md#swc

I don't really have a solution right now. I need to think more about it but a few ideas:
1) We could use undefined, we there is an indirection
2) We could fallback on a given type with a warning
3) We could look at the point location to refine the heuristic

PR #163, opened 2020-02-20
Exposing group interface to allow single-file morphologies loading

st4rl3ss:
This PR is part of a bigger operation that involves also `morpho-kit` and `TouchDetector`. The purpose of the whole operation is the introduction of a `Storage` object to allow a more flexible managing of morphologies loading. In particular the `Storage` object will make it easier to target morphologies that are in the current format (single directory with multiple files) while also introducing the possibility to load single-file morphologies. This new method of morphology loading would merge all the single .h5 files in a big h5 database, with the single files integrated as groups. Performance-wise the results were extremely promising in the prototyping phase. 

In order to allow single file loading, exposing the Group interface in MorphIO is necessary and this PR focuses on this, with some refactoring and rewriting of the error messages necessary since there's no explicit way in `HighFive` to export the name of a Group.

This PR is in no way to be considered definitive and it's meant to spark discussions on the matter.
If you want you can find the modification to `morpho-kit` and `TouchDetector` in the sandbox/bellotta/storage branch in both repos. The whole circuit building stack has been already tested as working in limited use cases.

### TO DO LIST:
- [x] Format Code
- [x] Fix Error Messages
- [x] Fix _getStructureDataSet exception method
- [x] Solve issues with CI
- [x] Solve Formatting Issues (Partially done, reverting changes to src/CMakeLists.txt)
- [x] Add tests for incorrect morphologies and group loading
- [x] Find and implement a better way to handle error messages in case of group loading


src/readers/morphologyHDF5.cpp
```diff
@@ -417,12 +427,11 @@ void MorphologyHDF5::_readPerimeters(int firstSectionOffset) {
 
     try {
         HighFive::SilenceHDF5 silence;
-        HighFive::DataSet dataset = _file->getDataSet(_d_perimeters);
+        HighFive::DataSet dataset = _group.getDataSet(_d_perimeters);
 
         auto dims = dataset.getSpace().getDimensions();
         if (dims.size() != 1) {
-            throw MorphioError("Reading morphology file '" + _file->getName() +
```
wizmer: What about setting the filename as a class attribute and setting it to the empty string (or something more meaningful like "HDF5 Group") when reading from a group ? This way we could keep the old error message.
I think it is an important feature to know on which file it failed when you are iterating over a very large number of morphologies and it suddenly fails.

st4rl3ss: Makes sense. I will work on it.

include/morphio/morphology.h
```diff
@@ -36,8 +37,9 @@ class Morphology
         Example:
             Morphology("neuron.asc", TWO_POINTS_SECTIONS | SOMA_SPHERE);
      */
-    explicit Morphology(const std::string& source, unsigned int options = NO_MODIFIER);
-    explicit Morphology(mut::Morphology);
+    Morphology(const std::string& source, unsigned int options = NO_MODIFIER);
+    Morphology(const HighFive::Group& group, unsigned int options = NO_MODIFIER);
+    Morphology(mut::Morphology);
```
matz-e: @st4rl3ss I think all these should still be explicit?

src/readers/morphologyHDF5.cpp
```diff
@@ -182,20 +190,24 @@ bool MorphologyHDF5::_readV2Metadata() {
...
+HighFive::DataSet MorphologyHDF5::_getStructureDataSet()
+{
     try {
         HighFive::SilenceHDF5 silence;
-        return _file->getDataSet(_d_structure);
+        return _group.getDataSet(_d_structure);
     } catch (const HighFive::DataSetException&) {
-        return _file->createDataSet<int>(_d_structure, HighFive::DataSpace({nSections, 3}));
+        throw;
+        //FIXME: NO EMPTY DATASET IS CREATED IN CASE THIS ONE IS EMPTY
```
matz-e: Please fix this, still. We may have to have a discussion about how this should be handled.

include/morphio/morphology.h
st4rl3ss: Thanks! Makes sense. Fixed it!

src/readers/morphologyHDF5.cpp
st4rl3ss: Fixed! Thanks for the help!

src/vasc/properties.cpp
```diff
@@ -65,8 +65,8 @@ static bool compare_section_structure(const std::vector<VascSection::Type>& vec1
     if (vec1.size() != vec2.size()) {
         if (verbose_)
             printError(Warning::UNDEFINED,
-                    "Error comparing " + name + ", size differs: " + std::to_string(vec1.size()) +
-                        " vs " + std::to_string(vec2.size()));
+                       "Error comparing " + name + ", size differs: " +
+                           std::to_string(vec1.size()) + " vs " + std::to_string(vec2.size()));
         return false;
```
pramodk: Curious about such formatting changes : is this result from different settings for formatting? or different clang-format?

st4rl3ss: I'm using the .clang_format file included in the project and I used CLION to reapply it to the whole project. Does it look off to you?

mgeplf: >  or different clang-format?

If they differ, that's the problem.  The included `ci/check_clang_format.sh` pins the clang-format at `9.0.0` at the moment, so differences show up if you're not using the same version.

st4rl3ss: Thanks for the clarification. I will investigate this.

st4rl3ss: I've checked and clang format 9.0.0 seems to be there. Are there guidelines on how you apply .clang-format in this project? or better, a pre-made script? Thanks!

wizmer: Running `./ci/check_clang_format.sh` should be everything you need.

wizmer:
Is there a draft specification somewhere of the HDF5 database that will have to be supported ? Should it be considered as an extension of the H5 format ?

matz-e:
> Is there a draft specification somewhere of the HDF5 database that will have to be supported ? Should it be considered as an extension of the H5 format ?

This would be more a task of MorphoKit, IMO. This here is more a generalization along the lines of `json.load` and `json.loads` in Python, for example. We don't need to define the database structure for MorphIO, as long as the "leaves" of the structure look exactly like the HDF5 files currently.

wizmer:
What would be nice would be to have the equivalent in the python binding to pass an h5py.Group to `Morphology`. But I don't know if there is a way to get a bridge between the HighFive and the h5py group ?
Maybe we don't need that now anyway...

tests/test_morphology.cpp
```diff
@@ -20,3 +20,7 @@ TEST_CASE("LoadNeurolucidaMorphology", "[morphology]") {
 
     REQUIRE(m.diameters().size() == 14);
 }
```
wizmer: Could you add a real test case where the new  constructor `Morphology(const HighFive::Group& group, unsigned int options = NO_MODIFIER)` is used

matz-e:
@st4rl3ss please bump the HighFive submodule and see if the build issues get resolved :)

tests/test_morphology.cpp
```diff
@@ -20,3 +22,13 @@ TEST_CASE("LoadNeurolucidaMorphology", "[morphology]") {
...
+
+TEST_CASE("LoadBadDimensionMorphology", "[morphology]") {
+    REQUIRE_THROWS(morphio::Morphology("data/h5/v1/monodim.h5"));
+}
+
+TEST_CASE("LoadMergedMorphology", "[morphology]") {
+    auto file = HighFive::File("data/h5/merged.h5", HighFive::File::ReadOnly);
+    REQUIRE_NOTHROW(morphio::readers::h5::MorphologyHDF5(
```
wizmer: Can you test any property of the loaded morphology, just to make sure the morphology is not empty ?
You can test `rootSections.size()` for example.

st4rl3ss: In order to test this I would need access to morphologies complete with the correct metadata, either in V1 or V2 versions that I would then manually merge. Do you know where I can find such morphologies? Thanks in advance!

wizmer: You can find some in `tests/data/h5/v1`

st4rl3ss:
please retest

wizmer:
The hangs seem to disappear when I comment `recursive-include 3rdparty *` in `MANIFEST.in`. I'm investigating.

matz-e:
> The hangs seem to disappear when I comment `recursive-include 3rdparty *` in `MANIFEST.in`. I'm investigating.

Thanks! I just flushed the caches in Travis, just for good measure, too. I'm wondering if we should exclude `.git` for the `3rdparty` directory?

tests/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ set_target_properties(unittests
   )
 
 target_link_libraries(unittests
-    PRIVATE morphio_shared
+    PRIVATE morphio_shared HighFive
```
wizmer: Why is this needed now ? Because of the HIGHFIVE_EXTERNAL ?

CMakeLists.txt
```diff
@@ -7,10 +7,14 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 
 option(BUILD_BINDINGS "Build the python bindings" ON)
 option(${PROJECT_NAME}_CXX_WARNINGS "Compile C++ with warnings" ON)
+option(EXTERNAL_HIGHFIVE "Use HighFive from external source" OFF)
 
 include(${PROJECT_SOURCE_DIR}/CMake/CompilerFlags.cmake)
 set(CMAKE_CXX_FLAGS "${FLAGS}")
 
+if (EXTERNAL_HIGHFIVE)
+  find_package(HighFive REQUIRED)
```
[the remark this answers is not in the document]
matz-e: I think we should have the following scenarios:

* Python builds: use submodule
* Builds for external libraries: use external H5, all software build should use the same version. Spack makes that easy for us… I don't think it's too hard doing this manually, either.

It's loosing the version pin, sure. We could fix that if you have a lower bound for HighFive that is required? Submodules pointing to `master` are not really an ideal reference either :-/

PR #165, opened 2020-03-03
HDF5 reader had a "_write" member, for no reason

mgeplf:


mgeplf:
> Interesting find.

I'm surprised no linters/warnings found it, tbh.

PR #167, opened 2020-03-05
Fix pylint error

wizmer:
by exposing public members directly in `__init__.py`


morphio/__init__.py
```diff
@@ -1 +1,41 @@
-from ._morphio import *
+from ._morphio import (
```
mgeplf: is this everything? (ie: equiv to '... import *')

If so (and I can't believe I'm going to write this), I think 'import * # noqa' makes sense in this case (and only this case), because the `pybind11` should only export what is necessary, right?

wizmer: I was not clear. The problem it fixes is `module morphio has no member blabla` when morphio is used in another module. The `from morphio import *` still had the problem.

It is importing *almost* everything it should. The star was importing also all enums in the global namespace (see the change in the test files), I think it should not have been the case in the first place.

wizmer: See for example https://bbpcode.epfl.ch/ci/job/nse.python.morph-tool.gerrit/321/TOXENV=lint,platform=rh7.x/console

mgeplf: icic

someday I want to track down why pylint doesn't find that stuff.

PR #168, opened 2020-03-06
Create LICENSE

wizmer:


haleepfl:
According to https://choosealicense.com/licenses/lgpl-3.0/ we should also include the GNU GPL license. So (as I read it) the LGPL text should be in the file named COPYING.LESSER and the GPL text in the file named COPYING.

PR #170, opened 2020-03-09
Refactor hdf5 read_points and avoid temp obj

ferdonline:


src/readers/morphologyHDF5.cpp
```diff
@@ -213,6 +213,35 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset) {
...
     auto& somaDiameters = _properties._somaLevel._diameters;
 
+    using float_2dvec = std::vector<std::vector<float>>;
+
+    auto loadPoints = [&](const float_2dvec& vec, bool has_neurites) {
+        const std::size_t section_offset = has_neurites
+                                               ? static_cast<std::size_t>(firstSectionOffset)
+                                               : vec.size();
+
+        // points and diameters are TrivialType's. Fastest to resize then assign values
```
matz-e:
```suggestion
        // points and diameters are PODs. Fastest to resize then assign values
```
just to get around the apostrophe + plural.

src/readers/morphologyHDF5.cpp
```diff
@@ -430,8 +411,9 @@ void MorphologyHDF5::_readPerimeters(int firstSectionOffset) {
                                                       perimeters.end());
     } catch (...) {
         if (_properties._cellLevel._cellFamily == FAMILY_GLIA)
-            throw MorphioError("No empty perimeters allowed for glia "
-                             "morphology");
+            throw MorphioError(
+                "No empty perimeters allowed for glia "
+                "morphology");
```
matz-e:
```suggestion
            throw MorphioError("No empty perimeters allowed for glia morphology");
```
Seems like some weird formatting stuff here…

src/readers/morphologyHDF5.cpp
```diff
@@ -451,15 +433,16 @@ void MorphologyHDF5::_read(const std::string& groupName,
...
 
         data.resize(dims[0]);
         dataset.read(data);
     } catch (...) {
         if (_properties._cellLevel._cellFamily == FAMILY_GLIA)
-            throw MorphioError("No empty perimeters allowed for glia "
-                               "morphology");
+            throw MorphioError(
+                "No empty perimeters allowed for glia "
+                "morphology");
```
matz-e:
```suggestion
            throw MorphioError("No empty perimeters allowed for glia morphology");
```
as above

src/readers/morphologyHDF5.cpp
```diff
@@ -213,6 +213,35 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset) {
     auto& somaPoints = _properties._somaLevel._points;
     auto& somaDiameters = _properties._somaLevel._diameters;
 
+    using float_2dvec = std::vector<std::vector<float>>;
```
wizmer: I'm not sure the alias is necessary considering you use it only once. Besides, `vec` could be called `hd5fData` or something that suggests this is the input data not the output ones.

wizmer: Even though new I think it was a mistake, MorphIO uses camel-case, so `has_neurites` -> `hasNeurites_`

src/readers/morphologyHDF5.cpp
```diff
@@ -199,10 +199,10 @@ HighFive::DataSet MorphologyHDF5::_getStructureDataSet(size_t nSections) {
 
 
 /**
-   Returns true if the neuron has no neurites
+   Returns true if the neuron has neurites (for MORPHOLOGY_VERSION_H5_2)
 **/
-static inline bool noNeurites(int firstSectionOffset) {
-    return (firstSectionOffset == -1);
+static inline bool v2HasNeurites(int firstSectionOffset) {
```
wizmer: Any reason for this change. I thought this was not specific to v2.

ferdonline: I noticed this function can only used in the context of v2 since in v1, according to the code, if there are no neurites the `firstSectionOffset` is gonna be positive, equal to the full length of the array.

matz-e: Seconded. It's also only used in code paths specific to v2.

PR #177, opened 2020-03-23
Point coordinates are written to files with 9 digits

wizmer:
This ensures they have have a unique float32 representation which is
identical to the float64 one.

### DIFF
--- a/src/mut/writers.cpp
+++ b/src/mut/writers.cpp
@@ -26,6 +26,8 @@ struct base_type {
 template <typename T>
 struct base_type<std::vector<T>>: base_type<T> {};
 
+constexpr int FLOAT_PRECISION_PRINT = 9;
+
 bool hasPerimeterData(const morphio::mut::Morphology& morpho) {
     return !morpho.rootSections().empty() && !morpho.rootSections().front()->perimeters().empty();
 }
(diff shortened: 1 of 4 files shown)


src/mut/writers.cpp
```diff
@@ -38,10 +40,11 @@ void writeLine(std::ofstream& myfile,
                float diameter) {
     using std::setw;
 
-    myfile << std::to_string(id) << setw(12) << std::to_string(type) << ' ' << setw(12)
-           << std::to_string(point[0]) << ' ' << setw(12) << std::to_string(point[1]) << ' '
-           << setw(12) << std::to_string(point[2]) << ' ' << setw(12)
-           << std::to_string(diameter / 2.f) << setw(12) << std::to_string(parentId) << '\n';
+    myfile << std::to_string(id) << setw(12) << std::to_string(type) << ' ' << setw(12);
```
mgeplf: use `FLOAT_PRECISION_PRINT`, no?

also, do you need a setw per output?

wizmer: On this line there are only `int`s.

> also, do you need a setw per output?

I've tried replacing the 2nd `setw` by a `' '` or by nothing, but it did not end-up being formatted correctly.

mgeplf: sigh, I read it wrong, my bad

PR #179, opened 2020-03-24
Add windows CI

wizmer:


include/morphio/shared_utils.tpp
```diff
@@ -1,8 +1,7 @@
-#include <cmath>
-
 #include <morphio/errorMessages.h>
 #include <morphio/types.h>
 
+# define PI           3.14159265358979323846f
```
mgeplf: https://stackoverflow.com/questions/6563810/m-pi-works-with-math-h-but-not-with-cmath-in-visual-studio
also, don't use defines for constatnts, pls

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,333 @@
...
+      fail-fast: false
+      matrix:
+        config:
+        - {
+            name: "Windows Latest MSVC", artifact: "Windows-MSVC.tar.xz",
+            os: windows-latest,
+            cc: "cl", cxx: "cl",
+            environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars64.bat"
+          }
+        # - {
```
mgeplf: can this be cleaned up?

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,333 @@
...
+    steps:
+    - uses: actions/checkout@v1
+
+    - uses: actions/checkout@v2
+    - name: Checkout submodules
+      shell: bash
+      run: |
+        # If your submodules are configured to use SSH instead of HTTPS please uncomment the following line
+        # git config --global url."https://github.com/".insteadOf "git@github.com:"
+        auth_header="$(git config --local --get http.https://github.com/.extraheader)"
```
mgeplf: I can't tell if this is from the example code, or if this is your own workaround, could you document it/link to where it comes from?

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,333 @@
...
+        file(TO_CMAKE_PATH "$ENV{GITHUB_WORKSPACE}/${cmake_dir}" cmake_dir)
+        message("::set-output name=cmake_dir::${cmake_dir}")
+
+        if (NOT "${{ runner.os }}" STREQUAL "Windows")
+          execute_process(
+            COMMAND chmod +x ninja
+            COMMAND chmod +x ${cmake_dir}/cmake
+          )
+        endif()
+        # message("::add-path::$ENV{GITHUB_WORKSPACE}")
```
mgeplf: please cleanup unnecessary stuff

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,333 @@
...
+        auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+        git submodule sync --recursive
+        git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+
+    - name: Download Ninja and CMake
+      id: cmake_and_ninja
+      shell: cmake -P {0}
+      run: |
+        set(cmake_version $ENV{CMAKE_VERSION})
+        set(ninja_version $ENV{NINJA_VERSION})
```
mgeplf: is ninja necessary?

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,333 @@
...
+          execute_process(
+            COMMAND chmod +x ninja
+            COMMAND chmod +x ${cmake_dir}/cmake
+          )
+        endif()
+        # message("::add-path::$ENV{GITHUB_WORKSPACE}")
+        # message("::add-path::${cmake_dir}")
+
+
+    - name: Download ccache
```
mgeplf: do we really need ccache?

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,333 @@
...
+
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.6' # Version range or exact version of a Python version to use, using SemVer's version range syntax
+        architecture: 'x64' # optional x64 or x86. Defaults to x64 if not specified
+
+    - name: Download HDF5
+      shell: bash
+      run: |
+        curl https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.10/hdf5-1.10.4/src/hdf5-1.10.4.zip -o hdf5.zip
```
mgeplf: version should be configurable earlier, possibly with a comment on why that version

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,333 @@
...
+        )
+        message("::set-env name=HDF5_ROOT::C:/Program Files/HDF_Group/HDF5/1.10.4")
+
+        if (NOT result EQUAL 0)
+          message(FATAL_ERROR "Bad exit status")
+        endif()
+
+        execute_process(COMMAND ccache -s)
+
+    - name: Check content
```
mgeplf: ?

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,333 @@
...
+      run: |
+        # If your submodules are configured to use SSH instead of HTTPS please uncomment the following line
+        # git config --global url."https://github.com/".insteadOf "git@github.com:"
+        auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+        git submodule sync --recursive
+        git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
+
+    - name: Download Ninja and CMake
+      id: cmake_and_ninja
+      shell: cmake -P {0}
```
mgeplf: why is cmake used everywhere as the scripting language? doesn't that make it extra hard to make work?

src/vasc/vasculature.cpp
```diff
@@ -18,7 +18,9 @@ Vasculature::Vasculature(const std::string& source) {
         throw UnknownFileType("File has no extension");
     }
 
-    if (access(source.c_str(), F_OK) == -1) {
+    // Cross-platform check of file existance
+    std::ifstream file(source.c_str());
+    if (!file) {
```
mgeplf: file might exist, you but you don't have permissions to read it, at which point the error msg is misleading

tests/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ set_target_properties(unittests
   )
 
 target_link_libraries(unittests
-    PRIVATE morphio_shared HighFive
+    PRIVATE morphio_static HighFive
```
mgeplf: why can't they be shared?

.github/workflows/build_cmake.yml
wizmer: It was taken from the official docs https://github.com/actions/checkout but apparently there is a better way now.

.github/workflows/build_cmake.yml
wizmer: Ninja seems to be the go for generator on windows from what I could understand. My knowledge is very limited though.

wizmer:
I will attempt a rewrite.

tests/CMakeLists.txt
wizmer: 'unittests' was failing to link with morphio lib. It was looking for 'morphio.lib' while the shared lib was called 'morphio.dll'. There is probably a better way to fix that but I'm A1 in CMAKE so I don't know.

tests/utils.py
```diff
@@ -35,9 +35,11 @@ def ignored_warning(warning):
 
 @contextmanager
 def _tmp_file(content, extension):
-    with tempfile.NamedTemporaryFile(suffix='.' + extension, mode='w') as tmp_file:
+    with tempfile.NamedTemporaryFile(suffix='.' + extension, mode='w', delete=False) as tmp_file:
         tmp_file.write(content)
         tmp_file.seek(0)
+        tmp_file.close()
```
wizmer: Windows can not use the file if it's still open.

ci/python_test.sh
```diff
@@ -4,21 +4,22 @@ set -euxo pipefail
 
 VENV=env
 
-rm -rf "$VENV"
-python3 -mvenv "$VENV"
+if [[ "$OSTYPE" != "msys" ]]; then
```
asanin-epfl, line 7: Why only for `msys` now? Before it was for all OS.

src/vector_utils.cpp
```diff
@@ -3,6 +3,8 @@
 #include <numeric>    // std::accumulate
 #include <sstream>    // std::stringstream
 #include <string>     // std::string
+#include <cctype>     // std::tolower
+#include <cmath>
```
asanin-epfl, line 7: They are not used anywhere. Why adding them?

ci/python_test.sh
wizmer: It's exept for `msys`, not only :)

src/vector_utils.cpp
wizmer: cctype is used for tolower, I don't remember for cmath but I'm pretty sure it is used by windows.

ci/python_test.sh
asanin-epfl: Yes, sorry. I'm useless here... :(

src/vector_utils.cpp
mgeplf: if it's windows only, please wrap it in an ifdef, so it's obvious why it's needed.  Also, adding the comments like I did for what is used where is really helpful in the future.

wizmer:
I have removed the unnecessary `<cmath>` header. I think we are good!

wizmer:
Unless objections, I will merge this by the end of the day.

.travis.yml
```diff
@@ -11,19 +12,35 @@ jobs:
...
 
+    #### MAC OS #######
     - os: osx
       osx_image: xcode10.1
       language: generic
       env:
         CXX: clang++
         CC: clang
+
+    #### WINDOWS #######
```
mgeplf, line 23: This header needs to be lower; the following stanza is still OSX.

Are the headers necessary, I usually find these sorts of things are prone to error.

.travis.yml
```diff
@@ -11,19 +12,35 @@ jobs:
...
+      env:
+        - MINCONDA_VERSION="latest"
+        - CMAKE_GENERATOR="Visual Studio 15 2017 Win64"
+        - TESTS_TARGET="RUN_TESTS";
+      install:
+        - choco install --yes miniconda3
+        - source C:/Tools/miniconda3/Scripts/activate
+        - conda config --set always_yes yes --set changeps1 no
+        - conda update -q conda
+        - conda install -c conda-forge hdf5
```
mgeplf, line 42: do the windows wheels work on a clean windows install, without conda?

ci/python_test.sh
mgeplf: I think the reason for msys should be described, so we know what it is, and why its there

tests/utils.py
mgeplf: the 'with' doesn't close the file?

.travis.yml
wizmer: True, I will remove them.

.travis.yml
wizmer: I haven't trying tackling the wheel problem yet. The goal of this PR is to enable Windows CI. It's not an official Windows support yet. I was planning to start working on the wheels in a subsequent PR.

tests/utils.py
wizmer: You're probably right. Let me try this.

wizmer: Actually, the `yield` happens within the `with` statement so the file is not closed yet.

wizmer:
Ping

wizmer:
I am merging this. I want to emphasize on the fact this is not providing Windows support, it is just testing the code under Windows. However this will help me going forward the Windows support and hopefully Windows wheels.

PR #181, opened 2020-03-26
Fix bug: h5 files were not sanitized when no modifier was passed

wizmer:
Sanitization now always happens.


asanin-epfl:
I think you would need a review from someone with more rights to this repo.

mgeplf:
Could you add a test?

I think we need more tests in general of this code, and not just relying on NeuroM to catch things.

wizmer:
I have already added the test ` test_single_children`. I think we are good.

mgeplf:
missed that; shouldn't it check w/ all options?

wizmer:
The test for Neurolucida is here
https://github.com/BlueBrain/MorphIO/blob/master/tests/test_2_neurolucida.py#L304

And for SWC, sanitization does nothing (there are no single child section, no duplicates issues) so I can't check it.

matz-e:
This seems like a major breaking change, but was not treated as such?

wizmer:
It depends who 'breaking' refers to. It's not breaking API-wise but yes it's an important change.

PR #182, opened 2020-04-02
cmake: install HighFive if we use the bundled version.

matz-e:
@wizmer this is required to have dependent libraries such as `morpho-kit` work *if* there is no external HighFive provided.


matz-e:
Sorry, already got a follow up, due to BlueBrain/HighFive#321

wizmer:
Np. Add me as reviewer when it's ready.

PR #186, opened 2020-05-06
NeuroLucida markers are parsed and kept in Morphology::markers

wizmer:
## Description

A marker is an [s-expression](https://fr.wikipedia.org/wiki/S-expression) at the top level of the Neurolucida file that contains additional information about the morphology. For example:

```lisp
("pia"
  (Closed)
  (MBFObjectType 5)
  (0 1 2 3)
  (3 4 5 4)
  (6 7 8 5)
  (9 10 11 6)
 )
```

This PR adds a `Morphology.markers` attribute that stores the markers found in the file.
A Marker object has 3 attributes:
- label
- points
- diameters.

## Specification

The following s-expressions are parsed:
- Any s-exp with a top level string. Like:
    ```lisp
    ("pia"
    (Closed)
    (MBFObjectType 5)
    (0 1 2 3)
    (3 4 5 4)
    (6 7 8 5)
    (9 10 11 6)
    )
    ```
- An sexp with one of the following top level s-exp:
  * Dot
  * FilledCircle
  * SnowFlake
  * Asterisk
  * OpenCircle
  * OpenStar
  * Flower
  * OpenSquare
  * FilledStar
  * DoubleCircle
  
  Example:
  ```lisp
  (FilledCircle
  (Color RGB (64, 0, 128))
  (Name "Marker 11")
  (Set "axons")
  ( -189.59    55.67    28.68     0.12)  ; 1
  )  ;  End of markers
  ```

## Context

This has been requested by @lidakanari in order to extract pia and layer markers from LNMC NeuroLucida files.
https://bbpteam.epfl.ch/project/issues/browse/CELLS-31

## Usage

```python
cell = Morphology(os.path.join(_path, 'pia.asc'))
all_markers = cell.markers
pia = m.markers[0]

# fetch the label marker with the `label` attribute
assert_equal(pia.label, 'pia')

# fetch the points with the `points` attribute
assert_array_equal(pia.points,
                       [[0, 1, 2],
                        [3, 4, 5],
                        [6, 7, 8],
                        [9, 10, 11]])

# fetch the diameters with the `diameters` attribute
assert_array_equal(pia.diameters, [3, 4, 5, 6])
```


tests/test_2_neurolucida.py
```diff
@@ -510,3 +510,29 @@ def test_markers():
...
+
+    assert_equal(len(m.markers), 2)
+    pia = m.markers[0]
+    assert_equal(pia.label, 'pia')
+    assert_array_equal(pia.points,
+                       [[0, 1, 2],
+                        [0, 2, 2],
+                        [0, 3, 2],
+                        [0, 4, 2]])
+    assert_array_equal(pia.diameters, [3, 3, 3, 3])
```
tristan0x, line 532: I am a bit paranoïd, but I would use distinct values instead of `[3, 3, 3, 3]`. Same for the points.

wizmer: Toujours aussi rapide le Tristan :) J'espere que le confinement se passe bien pour toi.
Done, I have changed the values.

binds/python/bind_misc.cpp
```diff
@@ -250,6 +250,19 @@ void bind_misc(py::module& m) {
...
 
+    py::class_<morphio::Property::Marker>(m,
+                                          "Marker",
+                                          "Container class for NeuroLucida extra Markers ")
+        .def_property_readonly(
+            "label", [](morphio::Property::Marker* a) { return a->_label; }, "Returns the label")
+        .def_property_readonly(
+            "points",
+            [](morphio::Property::Marker* a) { return a->_pointLevel._points; },
+            "Returns the list of coordinates of annotated points")
```
asanin-epfl, line 261: Returns the list of coordinates of markers?

binds/python/bind_misc.cpp
```diff
@@ -250,6 +250,19 @@ void bind_misc(py::module& m) {
...
+        .def_property_readonly(
+            "diameters",
+            [](morphio::Property::Marker* a) { return a->_pointLevel._diameters; },
+            "Returns the list of diameters of annotated points");
```
asanin-epfl, line 265: Returns the list of diameters of markers?

binds/python/bind_misc.cpp
```diff
@@ -250,6 +250,19 @@ void bind_misc(py::module& m) {
             [](morphio::Property::Annotation* a) { return a->_points._perimeters; },
             "Returns the list of perimeters of annotated points");
 
+    py::class_<morphio::Property::Marker>(m,
+                                          "Marker",
+                                          "Container class for NeuroLucida extra Markers ")
+        .def_property_readonly(
+            "label", [](morphio::Property::Marker* a) { return a->_label; }, "Returns the label")
```
asanin-epfl, line 257: use `Marker* m` instead of `Marker* a`?

binds/python/bind_misc.cpp
```diff
@@ -250,6 +250,19 @@ void bind_misc(py::module& m) {
...
+        .def_property_readonly(
+            "points",
+            [](morphio::Property::Marker* a) { return a->_pointLevel._points; },
```
asanin-epfl, line 260: use `Marker* m` instead of `Marker* a`?

binds/python/bind_misc.cpp
```diff
@@ -250,6 +250,19 @@ void bind_misc(py::module& m) {
...
+            "Returns the list of coordinates of annotated points")
+        .def_property_readonly(
+            "diameters",
+            [](morphio::Property::Marker* a) { return a->_pointLevel._diameters; },
```
asanin-epfl, line 264: use `Marker* m` instead of `Marker* a`?

tests/data/pia.asc
```diff
@@ -0,0 +1,38 @@
...
+  (3 4 5 4)
+  (6 7 8 5)
+  (9 10 11 6)
+ )
+
+("layer1-2"
+  (Color RGB (255, 251, 255))
+  (GUID "11594A5D0D7646B580E877CB40957E07")
+  (MBFObjectType 5)
+  (Resolution 1.846036)
```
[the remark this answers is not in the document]
wizmer: Sorry, I don't understand your proposal.

include/morphio/properties.h
```diff
@@ -155,6 +155,11 @@ struct Annotation {
     std::string _details;
 };
 
+struct Marker {
+    PointLevel _pointLevel;
+    std::string _label;
+};
+
 struct CellLevel {
```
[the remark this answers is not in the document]
wizmer: Very very good point. Will do !

tests/data/pia.asc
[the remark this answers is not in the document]
asanin-epfl: I mean those fields:
```
  (GUID "11594A5D0D7646B580E877CB40957E07")
  (MBFObjectType 5)
  (Resolution 1.846036)
```
Why are they here? Should they be loaded as well? I see them for the first time in morphologies.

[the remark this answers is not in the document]
wizmer: Me too. It's from a file Lida gave me. Maybe one day they will be parsed but for the time being, there is no need.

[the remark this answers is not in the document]
wizmer: It could be some addition in latest NeuroLucida releases.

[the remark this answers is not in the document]
asanin-epfl: ok then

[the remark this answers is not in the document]
lidakanari: There is a variety of info in the raw files that we get from reconstructors, so this is just a subset of the info saved in the file. However, at the moment we don't need to load the rest of them because they differ between version, reconstructor and lab so it will be hard to capture all possible combinations.

asanin-epfl:
If Mike or Tristan don't have any objections, I'm going to approve this.

mgeplf:
Need to look a this.

binds/python/bind_misc.cpp
```diff
@@ -250,6 +250,21 @@ void bind_misc(py::module& m) {
             [](morphio::Property::Annotation* a) { return a->_points._perimeters; },
             "Returns the list of perimeters of annotated points");
 
+    py::class_<morphio::Property::Marker>(m,
+                                          "Marker",
+                                          "Container class for NeuroLucida extra Markers ")
+        .def_property_readonly(
+            "label",
+            [](morphio::Property::Marker* marker) { return marker->_label; },
```
mgeplf, line 258: since these are bare pointers, are the lifetimes being respected?  What happens in a morphology object has it's refcount decremented, will it live long enough that the Properties still exist?

tests/data/pia.asc
```diff
@@ -0,0 +1,38 @@
+;	V3 text file written for MicroBrightField products.
```
mgeplf, line 1: have you run this over the full collection of ASC morphologies? Did the expected marker count match?

wizmer: I have checked a few morphologies but I don't know what is the expected number of markers in all the morphologies we have.

mgeplf: Could you grep for the markers, count the number of hits, and compare that to how many are found by the code?

wizmer: Markers are not easily greppable. They are just sexp that do not have a (Axon), (Cellbody) or (Dendrite) in the header. I can try to grep for strings and filter out the one we know (ie "CellBody") and see if it matches.

mgeplf: I just spot checked a couple files, but they seem to have a comment saying ';  End of markers', so I thought that might work to grep on?

wizmer: Not necessarily. I've run the following code to make sure that all markers where being found in our morphology repository. Markers are either top level sexps starting with a quote (except Cell Body quotes), or supported NeuroLucida markers. 

```python
from pathlib import Path
import yaml
import re

from tqdm import tqdm
from morphio import Morphology, set_maximum_warnings
from numpy.testing import assert_equal

def is_marker(line):
    return any((re.match('\("', line) and '"CellBody' not in line and '"Cell Body' not in line,
                re.match('\((Dot|FilledCircle|SnowFlake|Asterisk|OpenCircle|OpenStar|Flower|OpenSquare|FilledStar|DoubleCircle)', line)))

def test_markers():
    set_maximum_warnings(0)
    folder = '/home/bcoste/workspace/MorphologyRepository'
    paths = list(Path(folder).rglob('*.ASC'))
    for p in tqdm(paths):
        if p.name in {'cr161027_A_idA.ASC'}:
            continue
        cell = Morphology(p)
        lines = list(filter(is_marker, open(p).readlines()))

        s = [marker.label for marker in cell.markers]
        msg = f'In {p}\n{s}\n VS \n{lines}'
        assert_equal(len(cell.markers), len(lines), msg)
```

arnaudon:
@adrien-berchet @lidakanari , this is what we could use to read addititonal metadata from .asc files from recosntructors, let's have a play with it next week, so we can merge it if it works as expected.

binds/python/bind_misc.cpp
wizmer: Good point. I have changed it so that it no longer use pointers but `const ref` instead.

arnaudon:
Maybe one thing to add, I have tried on a .asc file which begins with:
```
;   V3 text file written for MicroBrightField products.
(Sections S1 "576985993_0003" 3 100 0
 S2 "576985993_0013" 103 100 0
 S3 "576985993_0023" 203 100 0
 S4 "576985993_0033" 303 100 0
 S5 "576985993_0043" 403 100 0
 S6 "576985993_0053" 503 100 0
```
and I think `Secttions` is not understood by Morphio, so I get a crash upon loading:
```
my_morph.asc:3:error
Unexpected token: 3
```

Would it be possible to ignore unknown sections in .asc files, as was done prior to this PR + print a warning saying which sections were ignored, in case they are useful?

Thanks!

arnaudon:
@wizmer , aka the best, it works, thanks! So if I load a morph with mut, and save it, it does not save the maker data, but we'll save it on our side. Would you recommand a best way to save it?

arnaudon (approved):

wizmer:
I could add also implement saving the marker when writing the mutable object if that's really necessary.

arnaudon:
no no, that's fine, we save this data somewhere else, so we don't carry it around in the morphology files, as we will not need it as it is, but apply some processing on it to extract placement information, etc... It's all good like that on my side!

arnaudon:
Just trying to save the marker data in pickle:
```
TypeError: can't pickle morphio._morphio.Marker objects
```
is it hard to make this picklable? I'll convert to dict for the moment.

wizmer:
It's far from trivial to get this right but here was the try: https://github.com/BlueBrain/MorphIO/pull/197

arnaudon:
OMG! 
I just did that:
```
   76 def convert_markers(markers):                                                                                                                                                                         
   77     """Convert morphio object to list of dicts."""                                                  
   78     _markers = []                                                                                   
   79     for marker in markers:                                                                          
   80         _marker = {}                                                                                
   81         _marker['label'] = marker.label                                                             
   82         _marker['points'] = marker.points                                                           
   83         _marker['diameters'] = marker.diameters                                                     
   84         _markers.append(_marker)                                                                    
   85     return _markers       
```
much simpler!

arnaudon:
shall we merge that before it t diverges to much?

PR #193, opened 2020-06-23
WIP: try getting hdf5 1.10.5

wizmer:
So that the hdf5 DLL shipped with morphio is compatible with the one shipped in h5py


wizmer:
retest this please

PR #194, opened 2020-08-12
I/O Improvements for HDF5 v1 Files

sergiorg-hpc:
Reading HDF5 v1 files showed unexpected 4-byte data accesses that clearly caused performance limitations inside TouchDetector. The reason for such a problem is due to an internal hyperslab selection and double-read of the `structure` dataset.

The changes of the pull request force the code to:

1. Read the whole dataset at once, instead of manually selecting the columns to read (which can be problematic for MPI-IO inside pHDF5).
2. Prevent reading the `structure` dataset twice, combining the reading of the types inside the same function that reads the sections.

Lastly, to confirm the differences in performance, **here is the output from Darshan _before_ the changes** (note the millions of 4-byte accesses):

https://www.dropbox.com/s/esk26eudlaeg1ry/srivas_h5benchmark_id179582_7-29-61389-1536220139145925938_2259662.darshan.pdf?dl=0

And **here is the output from Darshan _after_ the changes** (note execution time, IOps, I/O bandwidth, and so on):

https://www.dropbox.com/s/hpkf62bwurplf01/srivas_h5benchmark_id281591_8-11-66230-18035594655826765581_3384239.darshan.pdf?dl=0

These results are single-process only and obtained with a mini-application created to reproduce the issue that @matz-e found inside TouchDetector.


mgeplf:
Excited to look at this, but super low on time.

mgeplf:
From the darshan reports (in case drop box disappears):
Old:
image
image

New:
image
image

mgeplf:
I see that it's a single file being opened, correct?  What happens when the morphs are in many small files?  Does POSIX/GPFS time dominate?

src/readers/morphologyHDF5.cpp
```diff
@@ -275,126 +274,109 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset) {
...
-        if (skipFirst) {
-            skipFirst = false;
-            continue;
+    types.reserve(vec.size() - 1);  // remove soma type
+
+    for (size_t i = 1; i < vec.size(); ++i) {
+        const auto& p = vec[i];
+        const auto& type = p[1];
+
+        if (type > SECTION_CUSTOM_START || type < 0) {
```
mgeplf, line 300: From clang:
```
/home/travis/build/BlueBrain/MorphIO/src/readers/morphologyHDF5.cpp:368:49: error: 
      result of comparison of unsigned enum expression < 0 is always false
      [-Werror,-Wtautological-unsigned-enum-zero-compare]
        if (type > SECTION_CUSTOM_START || type < 0) {
                                           ~~~~ ^ ~
```
Prev. types was getting cast to int, which seems like a bug: https://github.com/BlueBrain/MorphIO/pull/194/files#diff-fecd1a78269c91432367e6141e692f06L378

I think @wizmer will have to look at that.

matz-e:
> I see that it's a single file being opened, correct? What happens when the morphs are in many small files? Does POSIX/GPFS time dominate?

This was the result of our initial benchmark:
```
                     time
                     mean count
version mode
5.5.1   db          223.3     3
        db_nested   230.3     3
        files        71.0     3
5.5.2   db          367.3     3
        db_nested   452.0     3
        files        86.0     3
        ime         159.7     3
        ime_nested  147.7     3
```

Using single files outperformed merged files by at least a factor of 3…

matz-e:
Got some new numbers. Definitely an improvement (only varying the data access):
```
                  HDF5    pHDF5     w/PR
nodes mode
4     db         108.0    216.5    134.2
      db_nested   99.5    247.5     83.2
      files       33.0     25.2     25.2
8     db          51.8    135.5     50.5
      db_nested   62.8    145.2     50.2
      files       24.5     18.0     19.0
16    db          50.2     66.0     39.0
      db_nested   43.8     80.8     32.8
      files       19.2     14.5     13.8
32    db          43.0     38.0     35.8
      db_nested   36.5     43.2     27.2
      files       17.5     10.2     10.2
```
Now storing everything in a single file is still lacking a bit compared to individual files, need to see about that.

src/readers/morphologyHDF5.cpp
sergiorg-hpc: This is strange, because at the end this is pointing to a traditional `enum` and the values can be negative numbers, right? Even replacing the `0` with `SECTION_UNDEFINED` (which is 0) might not work if the compiler is assuming unsigned values, strange!

mgeplf: `traditional enum` are implementation defined, it seems; so in this case clang chose `unsigned` it seems.

I think the logic of this needs to be checked; paging @wizmer.

sergiorg-hpc: If the files are expected to contain negative values for the types, I believe there is a way to enforce "typed enums", so that you can tell the compiler that it is a signed integer. But this is a C++11 feature I think.

mgeplf:
Sorry about the delay, @wizmer is on vacation, and I think it's important that he looks at this.

src/readers/morphologyHDF5.cpp
wizmer: I don't manage to reproduce the clang error message on my machine.
I also don't understand how clang comes to the conclusion that `type` is an enum given that:
```cpp
        std::vector<std::vector<int>> vec;
        const auto& p = vec[i];
        const auto& type = p[1];
```
So `type` should be an `int` in the first place. Actually @sergiorg-hpc , could you replace the two `auto` by their actual type, that will make things more explicit ? Maybe it's the comparison to `SECTION_CUSTOM_START` that is messing up with clang ?

But yes, I think the code is correct. The idea is to fill `types` with enum elements and to compare the integer value of each enum to see if they are out of range. I have added a test for negative values in this PR: https://github.com/BlueBrain/MorphIO/pull/195

wizmer (commented):
I think you should add a unit test on the C++ side that uses mocks to assert the number of calls to `read()`.
Otherwise there is the risk that one day someone will attempt to simplify the code, will merge back the v1 and v2 implems and will introduce a regression.

matz-e:
> I think you should add a unit test on the C++ side that uses mocks to assert the number of calls to read().

Got any frameworks for that in mind?

wizmer:
I was thinking of something like https://github.com/google/googletest/blob/master/googlemock/docs/cheat_sheet.md#using-mocks-in-tests-usingmocks
I know mocks in C++ are a massive pain so it's up to you to decide guys. 
An alternative would be to improve the MorphIO CI to run some benchmarks at every commits. Do you have something like this on the HPC side ?

src/readers/morphologyHDF5.cpp
sergiorg-hpc: I think the comparison is the actual problem. To give it a try, I just added a cast to `int` for the `SECTION_CUSTOM_START`, making the comparison to plain integers instead of `enum` types. This will maintain the original logic of the source code, without additional changes.

mgeplf: > I don't manage to reproduce the clang error message on my machine.

It's in the log (clang 7)
https://travis-ci.com/github/BlueBrain/MorphIO/jobs/379841972#L446

> I also don't understand how clang comes to the conclusion that type is an enum given that:
> std::vector<std::vector<int>> vec;

Shouldn't the type of that vector be something more specific than an `int`?

sergiorg-hpc: Perhaps @matz-e can elaborate this, but as far as I understand it, the `structure` dataset inside the morphology files contain integer values. This is just to read the file, then it is converted to the specific datatype.

However, the trick of casting to `int` didn't work. The compiler is assuming that `0` is somehow unsigned.

sergiorg-hpc: @wizmer's suggestion of forcing the type to `int` seemed to work, thanks a lot! Closing down this thread.

sergiorg-hpc:
@mgeplf / @wizmer, the CI is still failing for some reason and I'm not sure this is related to the changes from the PR:

https://travis-ci.com/github/BlueBrain/MorphIO/jobs/379893746

Do you have any advice?

wizmer:
> @mgeplf / @wizmer, the CI is still failing for some reason and I'm not sure this is related to the changes from the PR:
> 
> https://travis-ci.com/github/BlueBrain/MorphIO/jobs/379893746
> 
> Do you have any advice?

It's not related to this, you can consider the CI passing. I'll have a look later.

What about my question on how to protect this code against regression ?

sergiorg-hpc:
> > @mgeplf / @wizmer, the CI is still failing for some reason and I'm not sure this is related to the changes from the PR:
> > https://travis-ci.com/github/BlueBrain/MorphIO/jobs/379893746
> > Do you have any advice?
> 
> It's not related to this, you can consider the CI passing. I'll have a look later.
> 
> What about my question on how to protect this code against regression ?

Oh sorry, I missed your comments earlier. If I understood you correctly, your point is to ensure that we don't make more than one call to `read()`, is this correct? If so, the main problem of the current version is that it was selecting a subset of the whole dataset, causing many million individual read operations internally in HDF5. The code was only calling `read()` twice, once per subset,  but this was causing many individual reads internally. Hence, the PR is simplifying the dataset access by reading everything directly.

Could you, please, elaborate what was your main concern about regression?

matz-e:
I think that there is not really a good way to protect against this on a package level, it's pretty deep in the stack, and implementation dependent, too. If we hadn't started using HDF5 with MPI I/O, this would have not been noticed at all, even though there is a measurable performance improvement.

I would think we should merge this for now, and revisit the need for a regression test later. The fact alone that we're reading into a `vector<vector<float>>` with HighFive (where the inner vectors are allocated about 60k times for some morphologies) is impacting performance, and no one seems to be bothered by this.

wizmer:
@sergiorg-hpc My main concern is if that this PR adds an improvement which is hidden in the implementation and there is no unit test that protect this improvement from being destroyed in the future (like someone revisiting the h5 reader in a few weeks/months/year).

But yes, we can consider the PR ready for merge if you guys are willing to.

matz-e:
> @sergiorg-hpc My main concern is if that this PR adds an improvement which is hidden in the implementation and there is no unit test that protect this improvement from being destroyed in the future (like someone revisiting the h5 reader in a few weeks/months/year).

This is a good concern. @sergiorg-hpc maybe we can add a comment pointing out where we read more than needed to avoid performance penalties? Other than that, I hold the opinion that doing a larger-scale regression test on BlueBrain 5 would be more appropriate, since we have full control over the hardware and upgrade cycle, ensuring a somewhat consistent performance expectation.

Adding a unit test for the performance seems a bit futile to me, as there are a lot of environmental influences beyond our control. Would we set certain thresholds for execution time? What happens then if Travis will get beefier hardware and regressions fail to trip up our threshold? We can go to great lengths of trying to determine if there is a correct low-level number of reads, but I don't think that the time investment for that is entirely justified.

> But yes, we can consider the PR ready for merge if you guys are willing to.

Thanks! Once HighFive is updated, we (@sergiorg-hpc) would like to tackle changing the reads from `vector<vector>` to `vector<array>` to reduce the number of memory allocations performed.

Could you please remind us why the points and diameter are kept in different structures? If this were a single structure, we could actually just read directly into the `properties::points` and avoid another temporary structure…

mgeplf:
> Could you please remind us why the points and diameter are kept in different structures? If this were a single structure, we could actually just read directly into the `properties::points` and avoid another temporary structure…

This came up when discussing the design, and the concensus was that having XYZ in one contiguous array was better for when working w/ other libraries: Eigen/glm/etc all have adapters for arrays like that, but if you make it XYZR, it makes it more complex.

sergiorg-hpc:
> @sergiorg-hpc My main concern is if that this PR adds an improvement which is hidden in the implementation and there is no unit test that protect this improvement from being destroyed in the future (like someone revisiting the h5 reader in a few weeks/months/year).
> 
> But yes, we can consider the PR ready for merge if you guys are willing to.

Thanks @wizmer! As suggested by @matz-e, I have added a clarification to prevent someone changing the code back to how it was earlier. I have also included a small description about why we use `int` instead of `auto` in the type comparisons.

If you're fine with the changes, then we can go ahead and merge this PR.

matz-e (approved):

sergiorg-hpc:
Thank you all for the amazing help and valuable comments!

mgeplf:
> Thank you all for the amazing help and valuable comments!

Thank you for doing the heavy lifting of finding this, and fixing it.  Very much appreciated.

sergiorg-hpc:
> > Thank you all for the amazing help and valuable comments!
> 
> Thank you for doing the heavy lifting of finding this, and fixing it. Very much appreciated.

It was mostly @matz-e's work, he deserves the credit 👌

matz-e:
Nah, it's you @sergiorg-hpc! I just did the prodding… :D

Thanks!

PR #197, opened 2020-09-14
morphio::Morphology supports serialization through pickle

wizmer:
https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support

This is a simplistic attempt at supporting Pickle.
The serialization simply returns a string with the NeuroLucida representation of
the morphology.

A new morphio::Morphology ctor that takes a stringstream argument has been
added. It is used for the deserialization.

### DIFF
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -146,7 +146,6 @@ void bind_immutable_module(py::module& m) {
             "- morphio.IterType.breadth_first (default)\n"
             "iter_type"_a = IterType::DEPTH_FIRST)
 
-        // Make class picklable
         .def(py::pickle(
             [](const morphio::Morphology& morphology) {  // __getstate__
                 // Return a tuple that fully encodes the state of the object
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -7,6 +7,7 @@
 #include <morphio/endoplasmic_reticulum.h>
 #include <morphio/enums.h>
 #include <morphio/mut/morphology.h>
+#include <morphio/mut/writers.h>
 #include <morphio/soma.h>
 #include <morphio/types.h>
 
@@ -143,7 +144,32 @@ void bind_immutable_module(py::module& m) {
             "can be passed:\n"
             "- morphio.IterType.depth_first (default)\n"
             "- morphio.IterType.breadth_first (default)\n"
-            "iter_type"_a = IterType::DEPTH_FIRST);
+            "iter_type"_a = IterType::DEPTH_FIRST)
+
+        // Make class picklable
+        .def(py::pickle(
+            [](const morphio::Morphology& morphology) {  // __getstate__
+                // Return a tuple that fully encodes the state of the object
+                if (!morphio::mut::writer::_shouldBeWritten(morphology))
+                    return py::make_tuple("");
+                std::ostringstream oss;
+                morphio::mut::writer::_asc_to_stream(morphology, oss);
+
+                return py::make_tuple(oss.str());
+            },
+            [](py::tuple t) {  // __setstate__
+                if (t.size() != 1)
+                    throw std::runtime_error("Invalid state!");
+
+                auto str = t[0].cast<std::string>();
+
+                if (str.empty())
+                    return morphio::Morphology(morphio::mut::Morphology());
+
+                /* Create a new C++ instance */
+                const std::stringstream ss(str);
+                return morphio::Morphology(ss);
+            }));
 
 
     py::class_<morphio::Mitochondria>(
--- a/include/morphio/mut/writers.h
+++ b/include/morphio/mut/writers.h
@@ -6,6 +6,8 @@ namespace writer {
 void swc(const Morphology& morphology, const std::string& filename);
 void asc(const Morphology& morphology, const std::string& filename);
 void h5(const Morphology& morphology, const std::string& filename);
+void _asc_to_stream(const Morphology& morphology, std::ostream& stream);
+bool _shouldBeWritten(const Morphology& morphology);
 }  // namespace writer
 }  // end namespace mut
 }  // end namespace morphio
--- a/src/morphology.cpp
+++ b/src/morphology.cpp
@@ -230,7 +230,7 @@ Property::Properties loadURI(const std::string& source, unsigned int options) {
             std::string input((std::istreambuf_iterator<char>(ifs)),
                               (std::istreambuf_iterator<char>()));
 
-            return readers::asc::load(input, source, options);
+            return readers::asc::load(std::move(input), source, options);
         }
         if (extension == ".swc" || extension == ".SWC")
             return readers::swc::load(source, options);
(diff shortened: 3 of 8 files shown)


include/morphio/mut/writers.h
```diff
@@ -6,6 +6,8 @@ namespace writer {
 void swc(const Morphology& morphology, const std::string& filename);
 void asc(const Morphology& morphology, const std::string& filename);
 void h5(const Morphology& morphology, const std::string& filename);
+void _asc_to_stream(const Morphology& morphology, std::ostream& stream);
```
wizmer, line 9: These two new functions are only used in the binding code. I don't know if I should keep the underscore or not.

src/morphology.cpp
```diff
@@ -47,6 +47,9 @@ Morphology::Morphology(const Property::Properties& properties, unsigned int opti
 Morphology::Morphology(const HighFive::Group& group, unsigned int options)
     : Morphology(readers::h5::load(group), options) {}
 
+Morphology::Morphology(const std::stringstream& stream, unsigned int options)
+    : Morphology(readers::asc::load(stream.str(), "Stream", options), options) {}
```
wizmer, line 51: This new ctor assume the stream is in NeuroLucida format.
That's probably not good enough. Maybe we should be able to pass the format as a parameter ?

src/readers/morphologyASC.cpp
```diff
@@ -257,8 +255,8 @@ class NeurolucidaParser
     ErrorMessages err_;
 };
 
-Property::Properties load(const std::string& uri, unsigned int options) {
-    NeurolucidaParser parser(uri);
+Property::Properties load(const std::string&& input, const std::string& uri, unsigned int options) {
+    NeurolucidaParser parser(std::move(input), uri);
```
wizmer, line 259: I don't understand why I need to use `std::move` here since input is already an rvalue reference and NeurolucidaParser first argument also expect an rvalue reference...

wizmer:
@mgeplf I've added you mainly for the design. @st4rl3ss mainly for the C++

The rationale behind this PR is that I need the Morphology class to support pickling to rebase NeuroM v2. Indeed, we've recently added some multiprocessing features to NeuroM (https://github.com/BlueBrain/NeuroM/commit/4f01b4cabee1f469bd5f7313c0efbc9854dedf27) and now I can't rebase `mut_morphio` branch of NeuroM on top of master.

binds/python/bind_immutable.cpp
```diff
@@ -143,7 +144,32 @@ void bind_immutable_module(py::module& m) {
             "can be passed:\n"
             "- morphio.IterType.depth_first (default)\n"
             "- morphio.IterType.breadth_first (default)\n"
-            "iter_type"_a = IterType::DEPTH_FIRST);
+            "iter_type"_a = IterType::DEPTH_FIRST)
+
+        // Make class picklable
```
mgeplf, line 149: not sure this comment is needed; `py::pickle` should give it away, imo

binds/python/bind_immutable.cpp
```diff
@@ -143,7 +144,32 @@ void bind_immutable_module(py::module& m) {
...
+        .def(py::pickle(
+            [](const morphio::Morphology& morphology) {  // __getstate__
+                // Return a tuple that fully encodes the state of the object
+                if (!morphio::mut::writer::_shouldBeWritten(morphology))
```
mgeplf, line 153: I wonder if all the serialization code should be moved to this file

include/morphio/mut/writers.h
mgeplf: Can we put them in the bindings, so they don't become part of the `official` interface?

binds/python/bind_immutable.cpp
```diff
@@ -143,7 +144,32 @@ void bind_immutable_module(py::module& m) {
...
+                    throw std::runtime_error("Invalid state!");
+
+                auto str = t[0].cast<std::string>();
+
+                if (str.empty())
+                    return morphio::Morphology(morphio::mut::Morphology());
+
+                /* Create a new C++ instance */
+                const std::stringstream ss(str);
+                return morphio::Morphology(ss);
```
mgeplf, line 171: `morphio::Morphology` assuming that it's an asc file when handed a `stringstream` seems dangerous.  Wouldn't the mirror of the serialization use the inverse of `_asc_to_stream`?

src/readers/morphologyASC.cpp
```diff
@@ -42,21 +42,18 @@ bool skip_sexp(size_t id) {
 class NeurolucidaParser
 {
   public:
-    explicit NeurolucidaParser(const std::string& uri)
+    explicit NeurolucidaParser(const std::string&& input, const std::string& uri)
```
mgeplf, line 45: `const` r-value seems weird to me

src/readers/morphologyASC.cpp
mgeplf: Assuming that `load`, if passed a `std:string` is `asc` seems like we'll be boxing ourself into a corner when we want to support swc, for instance.

binds/python/bind_immutable.cpp
```diff
@@ -143,7 +144,32 @@ void bind_immutable_module(py::module& m) {
...
+            "iter_type"_a = IterType::DEPTH_FIRST)
+
+        // Make class picklable
+        .def(py::pickle(
+            [](const morphio::Morphology& morphology) {  // __getstate__
+                // Return a tuple that fully encodes the state of the object
+                if (!morphio::mut::writer::_shouldBeWritten(morphology))
+                    return py::make_tuple("");
+                std::ostringstream oss;
+                morphio::mut::writer::_asc_to_stream(morphology, oss);
```
mgeplf, line 156: I'm concerned we're using ASC for the serialization, since it can't capture what `SWC` encodes, which means that we potentially run problems.

Also, how does this handle `mitochondria` and `endoplasmicReticulum`?

src/mut/writers.cpp
```diff
@@ -163,20 +163,26 @@ static void _write_asc_section(std::ofstream& myfile,
     }
 }
 
-void asc(const Morphology& morphology, const std::string& filename) {
+/*
+ * Returns true if the morphology satisfies requirements to be written to disk in ASC format
+ * Returns false if the morphology is empty
+ * Raises if the morphology has perimeter data
+ */
+bool _shouldBeWritten(const Morphology& morphology) {
```
mgeplf, line 171: `should` -> `can`?

What's wrong with writing an empty morphology?

wizmer: The decision was made here https://github.com/BlueBrain/MorphIO/pull/128
But it's true that we could allow writing (and in this case, also reading) empty morphs. That would remove the need for this weird _shouldBeWritter function.
If we want to do this, we need to define what is a valid empty morph for each of the file formats.
Is it a totally empty file ? A file with a structure (empty dataset in the H5 case, a pair of () brackets in the ASC case ?).

binds/python/bind_immutable.cpp
wizmer: It does not handle stuff beyond xyzr. If we want to do this we need to use another approach.
I've opted for using ASC as the serialization medium for its simplicity but maybe I should already start digging further.

mgeplf: > [...] Indeed, we've recently added some multiprocessing features to NeuroM ([BlueBrain/NeuroM@4f01b4c]

Pickling/serialization is always hard.  I"m wondering if it would be possible to change the multiprocessing to not need pickling - only pass around filenames?

This PR could then be put off, since I think it will become more complicated than its current form.

PR #199, opened 2020-09-21
highfive: bump, remove bogus constructor.

matz-e:
Recent HighFive API changes block compilation with an external HighFive.
Update the HighFive submodule and remove legacy constructors in the HDF5
reader that broke with the API update.

### DIFF
--- a/src/readers/morphologyHDF5.h
+++ b/src/readers/morphologyHDF5.h
@@ -19,9 +19,6 @@ Property::Properties load(const HighFive::Group& group);
 class MorphologyHDF5
 {
   public:
-    MorphologyHDF5(const std::string& uri)
-        : _err(uri)
-        , _uri(uri) {}
     MorphologyHDF5(const HighFive::Group& group);
     virtual ~MorphologyHDF5() = default;
     Property::Properties load();
--- a/src/readers/morphologyHDF5.cpp
+++ b/src/readers/morphologyHDF5.cpp
@@ -61,10 +61,6 @@ namespace morphio {
 namespace readers {
 namespace h5 {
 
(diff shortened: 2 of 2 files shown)


matz-e:
I see only the Windows CI failing:
```

  File "c:\tools\miniconda3\lib\site-packages\nose\case.py", line 198, in runTest

    self.test(*self.arg)

  File "C:\Users\travis\build\BlueBrain\MorphIO\tests\test_6_writers.py", line 77, in test_write_basic

    import h5py

  File "c:\tools\miniconda3\lib\site-packages\h5py\__init__.py", line 34, in <module>

    from . import version

  File "c:\tools\miniconda3\lib\site-packages\h5py\version.py", line 17, in <module>

    from . import h5 as _h5

  File "h5py\h5.pyx", line 1, in init h5py.h5

ImportError: DLL load failed while importing defs: The specified procedure could not be found.
```

Seems unrelated?

wizmer (approved):

src/readers/morphologyHDF5.h
```diff
@@ -19,9 +19,6 @@ Property::Properties load(const HighFive::Group& group);
 class MorphologyHDF5
 {
   public:
-    MorphologyHDF5(const std::string& uri)
-        : _err(uri)
-        , _uri(uri) {}
```
wizmer, line 24: @matz-e I just realized that this commit broke the `uri` which no longer contains the morphology filename. This is espicially useful in case MorphIO fails because it used to tell you which file failed. Do you know what was the issue with this ctor ?

matz-e: IIRC, that `HighFive::Group` has no default constructor…

Actually, recovering the filename should be possible with HighFive now, so maybe this could be reworked.

PR #201, opened 2020-10-29
performance: nested vectors are bad

matz-e:
HighFive will allocate every inner vector, and copy data into them.
Just for us to copy the data out of the temporary nested vector. Avoid a
bit of copy by just providing a slab of memory.

### DIFF
--- a/src/readers/morphologyHDF5.cpp
+++ b/src/readers/morphologyHDF5.cpp
@@ -27,7 +27,7 @@
 namespace {
 // v1 & v2
 const std::string _d_points("points");
-const size_t _pointColumns = 4;
+constexpr size_t _pointColumns = 4;
 
 // v1
 const std::string _d_structure("structure");
@@ -216,7 +216,7 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset) {
     auto& somaPoints = _properties._somaLevel._points;
     auto& somaDiameters = _properties._somaLevel._diameters;
 
-    auto loadPoints = [&](const std::vector<std::array<float, 4>>& hd5fData, bool hasNeurites) {
+    auto loadPoints = [&](const std::vector<std::array<float, _pointColumns>>& hd5fData, bool hasNeurites) {
         const std::size_t section_offset = hasNeurites ? std::size_t(firstSectionOffset)
                                                        : hd5fData.size();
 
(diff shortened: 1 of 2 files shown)


src/readers/morphologyHDF5.cpp
```diff
@@ -138,12 +138,6 @@ void MorphologyHDF5::_resolveV1() {
     }
 
     _sections.reset(new HighFive::DataSet(_group.getDataSet(_d_structure)));
-    dataspace = _sections->getSpace();
-    _sectionsDims = dataspace.getDimensions();
-    if (_sectionsDims.size() != 2 || _sectionsDims[1] != _structureV1Columns) {
-        throw morphio::RawDataError("Error opening morphology:" + _uri +
-                                    " bad number of dimensions in 'structure' dataspace.");
-    }
```
matz-e, line 146: These checks are now done on line 290-ish

src/readers/morphologyHDF5.cpp
```diff
@@ -289,8 +288,17 @@ int MorphologyHDF5::_readV1Sections() {
...
+    const auto dims = _sections->getSpace().getDimensions();
+
+    if (dims.size() != 2 || dims[1] != _structureV1Columns) {
+        throw(MorphioError("Error reading morphologies " + _uri +
+                           " bad number of dimensions in 'structure' dataspace"));
+    }
+
+    std::vector<std::array<int, _structureV1Columns>> vec(dims[0]);
+    if (vec.size() > 0) {
+        _sections->read(vec.front().data());
```
mgeplf, line 300: This strikes me as dangerous. `data()` returns a bare pointer; are we sure the allocated size will be enough no matter what? Because `_sections->read()` can't do any bounds-checking (which I assume it did before?)

mgeplf:
I think it's best of @wizmer reviews this; I'm not familiar enough w/ the codebase.

src/readers/morphologyHDF5.cpp
matz-e: I don't like it either, but it's the same logic essentially, just shifted from HighFive to here and avoiding a bunch of inner vector allocations. There's a dimension check before this…

mgeplf: The HighFive logic is evolving, and should become better over time, no? By doing it ourselves, we have to backport that.

Can't HighFive do a `vec.reserve()` to avoid the inner vec alloations?

> There's a dimension check before this…

Are type sizes considered (along w/ the widening narrowing)?

matz-e: HighFive should become better, yes. But that may still be off for a while.

> Can't HighFive do a `vec.reserve()` to avoid the inner vec alloations?

On the outer vector yes, but that leaves individual allocations for thousands of inner vectors.

mgeplf: > On the outer vector yes, but that leaves individual allocations for thousands of inner vectors.

Ah, it can't use std::array like is done here?

matz-e: Not right now.

matz-e: FWIW, I _think_ this is the logic to get HDF5 to read with the right datatype, even if we just pass in a ptr:

https://github.com/BlueBrain/HighFive/blob/4c70d818ed18231563fe49ff197d1c41054be592/include/highfive/bits/H5Slice_traits_misc.hpp#L183

wizmer:
LGTM!

wizmer (approved):

src/readers/morphologyHDF5.cpp
mgeplf: I think there should be comment that reminds us to go back to the (safer, imo) way once HighFive does multi-dim with std::array.

PR #203, opened 2020-11-10
Try building the wheel with cibuildwheel

wizmer:


tomdele (approved):

.github/workflows/upload-wheels.yml
```diff
@@ -0,0 +1,81 @@
+name: Build
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+
+env:
+  CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
```
mgeplf, line 10: can this be done w/ manylinux2010?

.github/workflows/upload-wheels.yml
```diff
@@ -0,0 +1,81 @@
...
+          python-version: '3.7'
+
+      - name: Install cibuildwheel
+        run: |
+          python -m pip install cibuildwheel==1.6.4
+
+      - name: Build wheels on Linux
+        if: runner.os == 'Linux'
+        env:
+          CIBW_BEFORE_BUILD: 'yum update -y; yum -y install hdf5-devel'
```
mgeplf, line 44: can you check the produced wheel to make sure it's using a static hdf5 version (and not a rpath versioned .so); otherwise we may run into problems with conflicts of people using different hdf5 versions.

Issue #162, opened 2020-02-19
Add support for double-precision floating-point 


alexsavulescu:
We would need to have the possibility of using Morphio with doubles, as well as its python bindings.  For example, NRN is using doubles for points and their computation. 

After discussing with @wizmer and @ferdonline, we could consider one of the following: 
- CMAKE option to enable double support
- just use double as default

TBD

mgeplf:
In the original design, space was a big concern: the viz team brought up the fact they keep many morphs in memory at once, and since the source data is of low precision (float in hdf5, for Neurolucida ASCII, the points only have 3 decimal points - plus these positions are not very precise in the first place), that doubling the space requirement wouldn't bring much to the table.

alexsavulescu:
What I'm working on is a way to use MorphIo for morphology loading in NRN. Since the library is parsing floats and then passing back numpy ndarray of floats, this impacts the soma points calculations so we end up with different results. True,these differences might seem insignificant as they slightly differ (i.e. -0.576182 vs -0.576181) but they will have an impact in the overall simulation run, leading to different number of spikes and results a bit off from what you would get with NRN morphology loading. 

Would you then agree to have a CMAKE option to enable double support?

mgeplf:
> Would you then agree to have a CMAKE option to enable double support?

I'm not against it, I'm just not sure it will solve your problem: like I said, h5 files store floats internally.  The ASCII files might give the same results.

Keeping the (already complex) library complexity down is also nice, so the changes would hopefully not be too invasive.

alexsavulescu:
I see your point. Only worked with ASCII so far, I have to check h5 out. 

The changes would not be too invasive :)

wizmer:
Let's summarize.

**1) Keeping float in MorphIO**

Plus: 
- Size stay minimal
- Less complexity

Minus:
NRN loader and MorphIO loader differ for ASC files


**2) CMake flag:**

Plus:
NRN loader and MorphIO loader can be made equivalent for ASC files

Minus:
- More complexity
- MorphIO will now have multiples outcome because of the flag.

Question:
Should the flag propagate to the python bindings ?

**3) Move MorphIO to double:**

Plus:
NRN and MorphIO now behaves exactly the same way.

Minus:
Double the size of the morphology object.

Note:
From what I have seen, Viz team would probably discard the morphology after fetching the points and the relevant information they need. They could probably cast back to float there if needed.

===============================================================
[...]

PR #207, opened 2020-11-13
Allow doubles via CMake option #162

alexsavulescu:
By using `-DUSE_DOUBLE=ON` MorphIO is using doubles instead of floats. Default value is `OFF`.

Btw, getting these errors via `nosetests` (even with `master`): 
```python
======================================================================
ERROR: test_0_API.test_doc_exists
----------------------------------------------------------------------
  File "/Users/savulesc/Workspace/MorphIO/venv/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/savulesc/Workspace/MorphIO/tests/test_0_API.py", line 15, in test_doc_exists
    morphio.mut.EndoplasmicReticulum,
AttributeError: module 'morphio.mut' has no attribute 'EndoplasmicReticulum'

======================================================================
ERROR: Both modules morphio and morphio.mut should expose the same API
----------------------------------------------------------------------
  File "/Users/savulesc/Workspace/MorphIO/venv/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/savulesc/Workspace/MorphIO/tests/test_0_API.py", line 47, in test_mut_immut_have_same_methods
    methods(morphio.mut.EndoplasmicReticulum))
AttributeError: module 'morphio.mut' has no attribute 'EndoplasmicReticulum'

----------------------------------------------------------------------

```


setup.py
```diff
@@ -49,6 +49,14 @@ def get_cpu_count():
 
 
 class CMakeBuild(build_ext):
+    user_options = build_ext.user_options + [
```
alexsavulescu, line 52: Will use this for spack recipe (variant `use_double`)

src/mut/writers.cpp
```diff
@@ -37,12 +37,17 @@ void writeLine(std::ofstream& myfile,
...
-               float diameter) {
+               morphio::floatType diameter) {
     using std::setw;
 
+
     myfile << std::to_string(id) << setw(12) << std::to_string(type) << ' ' << setw(12);
-    myfile << std::fixed << std::setprecision(FLOAT_PRECISION_PRINT) << point[0] << ' ' << setw(12)
-           << point[1] << ' ' << setw(12) << point[2] << ' ' << setw(12) << diameter / 2.f
+    myfile << std::fixed
+#if !defined(USE_DOUBLE)
```
alexsavulescu, line 46: This was used to get as close as possible to doubles.

binds/python/bind_mutable.cpp
```diff
@@ -277,7 +277,8 @@ void bind_mutable_module(py::module& m) {
             "at each root section",
             "section_id"_a = -1);
 
-    using mitosection_floats_f = std::vector<float>& (morphio::mut::MitoSection::*) ();
+    using mitosection_floatTypes_f = std::vector<morphio::floatType>& (
```
matz-e, line 280: over-eager `s///g`?
```suggestion
    using mitosection_floats_f = std::vector<morphio::floatType>& (
```

CMakeLists.txt
```diff
@@ -8,6 +8,14 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 option(BUILD_BINDINGS "Build the python bindings" ON)
 option(${PROJECT_NAME}_CXX_WARNINGS "Compile C++ with warnings" ON)
 option(EXTERNAL_HIGHFIVE "Use HighFive from external source" OFF)
+option(USE_DOUBLE "Use doubles instead of floats" OFF)
```
matz-e, line 11: I feel like the project name should be added to the configuration option. Seems pretty generic otherwise.

include/morphio/properties.h
```diff
@@ -28,10 +28,11 @@ namespace Property {
...
 extern template void _appendVector(std::vector<Point>& to, const std::vector<Point>& from, int
-offset); extern template void _appendVector(std::vector<float>& to, const std::vector<float>& from,
-int offset); extern template void _appendVector(std::vector<unsigned int>& to, const
-std::vector<unsigned int>& from, int offset); template <typename T> std::vector<typename T::Type>
-copySpan(const std::vector<typename T::Type>& data, SectionRange range);
+offset); extern template void _appendVector(std::vector<morphio::floatType>& to, const
+std::vector<morphio::floatType>& from, int offset); extern template void
+_appendVector(std::vector<unsigned int>& to, const std::vector<unsigned int>& from, int offset);
+template <typename T> std::vector<typename T::Type> copySpan(const std::vector<typename T::Type>&
+data, SectionRange range);
```
matz-e, line 35: unnecessary reformat?

include/morphio/vector_types.h
```diff
@@ -7,14 +7,20 @@
 #include <vector>
 
 namespace morphio {
-using Point = std::array<float, 3>;
+#ifdef USE_DOUBLE
+using floatType = double;
```
matz-e, line 11: what about `long double`?

src/properties.cpp
```diff
@@ -143,7 +143,7 @@ bool compare(const morphio::range<const morphio::Point>& vec1,
         return false;
     }
 
-    const float epsilon = 1e-6f;
+    const floatType epsilon = static_cast<morphio::floatType>(1e-6);
```
matz-e, line 146: One thing I'm wondering about these conversions… if the default type is double, are resulting floats bit-identical. Same for the reverse scenario? What about not doing these casts? Compiler warnings that are over-zealous?

src/readers/morphologyHDF5.cpp
```diff
@@ -216,32 +216,33 @@ void MorphologyHDF5::_readPoints(int firstSectionOffset) {
...
-        }
-
-        if (hasNeurites) {
-            const size_t size = (points.size() + hd5fData.size() - section_offset);
-            points.resize(size);
-            diameters.resize(size);
-            for (std::size_t i = section_offset; i < hd5fData.size(); ++i) {
+    auto loadPoints =
+        [&](const std::vector<std::array<morphio::floatType, _pointColumns>>& hd5fData,
+            bool hasNeurites) {
```
matz-e, line 221:
```suggestion
    auto loadPoints = [&](
        const std::vector<std::array<morphio::floatType, _pointColumns>>& hd5fData,
        bool hasNeurites
    ) {
```
Embrace it… then keep indentation as before.

src/readers/morphologySWC.cpp
```diff
@@ -198,11 +198,11 @@ class SWCBuilder
...
-        float x = root.point[0];
-        float y = root.point[1];
-        float z = root.point[2];
-        float d = root.diameter;
-        float r = root.diameter / 2.f;
+        floatType x = root.point[0];
+        floatType y = root.point[1];
+        floatType z = root.point[2];
+        floatType d = root.diameter;
+        floatType r = root.diameter / static_cast<morphio::floatType>(2);
```
matz-e, line 205: By now there could be a global
```
constexpr morphio::floatType TWO = 2
```
wherever you define `floatType`, and this would read a little nicer. Or use Roman numerals.

matz-e (commented):
TIHI.

Left some nitpicky comments. Will all this make neuron numbers match to the internal parser?

binds/python/bind_mutable.cpp
alexsavulescu: indeed

CMakeLists.txt
alexsavulescu: Will add for more clarity. Thanks

include/morphio/properties.h
alexsavulescu: clang-format script from the repo (enforced by CI)

include/morphio/vector_types.h
alexsavulescu: NEURON uses double. But if needed one can adjust the code once again :D

src/properties.cpp
alexsavulescu: Yes, needed to keep compilers happy. better than adding macros. I will test with user-defined literals

alexsavulescu:
> TIHI.
> 
> Left some nitpicky comments. Will all this make neuron numbers match to the internal parser?

Thank you for the CR. Numbers should match, given that ASC and H5 have the same values. That's why I took out the precision when writing out morphs for doubles. IIRC, from original ASCII we get H5 and then write back ASCII.

src/properties.cpp
matz-e: Uh, that sounds like a good idea! Keep in mind that we're handicapped by the f'in wheel-building process, though.

wizmer:
In https://github.com/BlueBrain/MorphIO/blob/master/.github/workflows/run-tests.yml#L39,
could you now call cpp_test.sh twice, one for the float build and one for the double ?

include/morphio/properties.h
alexsavulescu: removed commented out code

src/mut/writers.cpp
```diff
@@ -37,13 +37,17 @@ void writeLine(std::ofstream& myfile,
...
     using std::setw;
 
+
     myfile << std::to_string(id) << setw(12) << std::to_string(type) << ' ' << setw(12);
-    myfile << std::fixed << std::setprecision(FLOAT_PRECISION_PRINT) << point[0] << ' ' << setw(12)
-           << point[1] << ' ' << setw(12) << point[2] << ' ' << setw(12) << diameter / 2.f
-           << setw(12);
+    myfile << std::fixed
+#if !defined(MORPHIO_USE_DOUBLE)
+           << std::setprecision(FLOAT_PRECISION_PRINT)
```
alexsavulescu, line 47: This was used to get as close as possible to doubles.

src/readers/morphologyHDF5.cpp
alexsavulescu: Nope

wizmer (approved):

matz-e (approved):

PR #208, opened 2020-11-16
Fix the wheel CI

wizmer:


wizmer:
@mgeplf `manylinux2010` does not have `cmake >= 3` which is required for morphio.
I can build cmake from sources and use manylinux2010 if you want.
My personal opinion is that 2014 is fine, all our machine at BBP are 2014 compliant. If someone comes here and complain, we can reconsider but chances are that it will never happen.

.github/workflows/publish-sdist-wheels.yml
```diff
@@ -45,7 +45,11 @@ jobs:
       - name: Build wheels Mac OS
         if: runner.os == 'macOS'
         env:
-          CIBW_BEFORE_BUILD: 'brew update; brew unlink gcc@8 gcc@9; brew install cmake hdf5'
+          CIBW_BEFORE_BUILD: |
```
mgeplf, line 48: is there a behavior change here?

wizmer: No, I think the only change was in brew itself.

mgeplf: k, thanks.

PR #214, opened 2020-12-03
Add windows wheels

wizmer:


wizmer:
The last commit is not building the wheels since I have switch back the wheel building trigger to "on tags".
The successful build of the wheels is in the previous commit: https://github.com/BlueBrain/MorphIO/runs/1511986706

PR #215, opened 2020-12-03
Add -P option to nosetests

wizmer:
Add \_\_init\_\_.py in tests folder

Also:
use pathlib instead of pathlib2

Kudos to Mike for the discover magic.


wizmer:
I have no idea of what I'm doing.

mgeplf:
What was the initial problem that this was trying to solve?  'Tests can run from any location' was the name of my PR in libsonata, b/c the `def test_suite` was hardcoding a path.

wizmer:
I had some issue in the CI wheel PR because the tests.utils module was not found so I had to add an \_\_init.py\_\_ in tests/ which in turn broke `ci/test_python.sh`.
The discover magic allows to both have \_\_init.py\_\_ in the tests folder and `ci/test_python.sh` working.

asanin-epfl:
In our Python projects we stay within the paradigm of `from utils import` not `from . utils import`. Is `.` usage necessary? Anyway, @mgeplf should say the final word here. I don't quite understand what's happening.

wizmer:
For some reason that I did not understand, the tests, when run as part of the Windows wheel building are not passing with the old way.
https://github.com/BlueBrain/MorphIO/runs/1493403378?check_suite_focus=true#step:9:697

I end-up with a lot of:
```
File "D:\a\MorphIO\MorphIO\tests\test_7_modifiers.py", line 8, in <module>
  from utils import captured_output
ModuleNotFoundError: No module named 'utils'
```
So I tried a bunch of stuff and this worked...

asanin-epfl:
How urgent is this?

wizmer:
https://github.com/BlueBrain/MorphIO/pull/214 will need to be rebased on top of this commit.
https://bbpteam.epfl.ch/project/issues/browse/NSETM-1018 is now on priority high so I guess somewhat urgent.

asanin-epfl (approved):

tests/__init__.py
```diff
@@ -0,0 +1,7 @@
+import unittest
+
+
+def test_suite():
+    loader = unittest.TestLoader()
+    suite = loader.discover('tests', pattern='test*.py')
```
mgeplf, line 6: isn't this hardcoding the directory?

mgeplf:
Looking at this more; I think the fix is the `-P` and the correct relative imports.  Can you test this? It identifies the root cause much more clearly.  Then, we need to figure out what they -P is needed.

wizmer:
I've just tested this. Without the `-P`, nosetests append the `src` folder at the beginning of $PATH. Since there is a morphio folder in src, it will find this one first instead of finding the true morphio package installed in site-packages.

mgeplf:
That seems a lot cleaner, thanks.

PR #221, opened 2021-01-15
Handling missing duplicate points

wizmer:
## What is a duplicate point ?

When a child section starts with a point at the same coordinates as the parent
section last point, we call this a 'duplicate point'.

ℹ️  To consider a point as a duplicate, we only consider its spatial coordinates.
The diameter is not taken into account. This means a duplicate is allowed to
have any diameter value.

## Context:

Some ASC files comes with duplicate points, some do not. So it seems that both
representations are valid and need to be supported.

## Specification

The following specification has been created to exactly recreate NEURON's behavior:

* When a duplicate point is **not** in the file (ie. a section does not start at the
coordinate of its parent section last point), the duplicate is added. The
diameter of the added point is the diameter of the point that comes right after
it.

* When a duplicate is already in the file, its coordinates and diameter value
are used as they are.

* When a single child section is merged with its parent section and the child
section has a duplicate with a different diameter value, the diameter of the
**parent** section is kept.

## More information

* https://bbpteam.epfl.ch/project/issues/browse/NSETM-1178?focusedCommentId=135030&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-135030
* https://github.com/BlueBrain/NeuroM/issues/654#issuecomment-332864540

### DIFF
--- a/src/readers/morphologyASC.cpp
+++ b/src/readers/morphologyASC.cpp
@@ -123,11 +123,10 @@ class NeurolucidaParser
             return_id = -1;
         } else {
             SectionType section_type = TokenSectionTypeMap.at(token);
-            insertLastPointParentSection(parent_id, properties);
+            insertLastPointParentSection(parent_id, properties, diameters);
 
-            // Condition to remove single point section that duplicate parent
-            // point See test_single_point_section_duplicate_parent for an
-            // example
+            // Condition to remove a single point section that is made of only the duplicate point
+            // See test_single_point_section_duplicate_parent for an example
             if (parent_id > -1 && properties._points.size() == 1) {
                 return_id = parent_id;
             } else {
@@ -152,32 +151,41 @@ class NeurolucidaParser
       Add the last point of parent section to the beginning of this section
       if not already present.
 
-     The idea is that these two structures should represent the same morphology:
+      The diameter is taken from the child section next point as does NEURON.
+      Here is the spec:
+      https://bbpteam.epfl.ch/project/issues/browse/NSETM-1178?focusedCommentId=135030&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-135030
 
-     (3 -8 0 2)     and          (3 -8 0 2)
-     (3 -10 0 2)                 (3 -10 0 2)
-     (                           (
-       (0 -10 0 2)                 (3 -10 0 2)  <-- duplicate parent point
-       (-3 -10 0 2)                (0 -10 0 2)
-       |                           (-3 -10 0 2)
-       (6 -10 0 2)                 |
-       (9 -10 0 2)                 (3 -10 0 2)  <-- duplicate parent point
-     )                             (6 -10 0 2)
-                                   (9 -10 0 2)
-                                 )
+      In term of diameters, the result should look like the right picture of:
+      https://github.com/BlueBrain/NeuroM/issues/654#issuecomment-332864540
+
+      The idea is that these two structures should represent the same morphology:
+
+      (3 -8 0 2)     and          (3 -8 0 2)
+      (3 -10 0 2)                 (3 -10 0 2)
+      (                           (
+        (0 -10 0 2)                 (3 -10 0 2)  <-- duplicate parent point
+        (-3 -10 0 2)                (0 -10 0 2)
+        |                           (-3 -10 0 2)
+        (6 -10 0 2)                 |
+        (9 -10 0 2)                 (3 -10 0 2)  <-- duplicate parent point
+      )                             (6 -10 0 2)
+                                    (9 -10 0 2)
+                                  )
      */
-    void insertLastPointParentSection(int32_t parentId, morphio::Property::PointLevel& properties) {
+    void insertLastPointParentSection(int32_t parentId,
+                                      morphio::Property::PointLevel& properties,
+                                      std::vector<morphio::floatType>& diameters) {
         if (parentId < 0)  // Discard root sections
             return;
         auto parent = nb_.section(static_cast<unsigned int>(parentId));
         auto lastParentPoint = parent->points()[parent->points().size() - 1];
-        auto lastParentDiameter = parent->diameters()[parent->diameters().size() - 1];
+        auto childSectionNextDiameter = diameters[0];
 
         if (lastParentPoint == properties._points[0])
             return;
 
         properties._points.insert(properties._points.begin(), lastParentPoint);
-        properties._diameters.insert(properties._diameters.begin(), lastParentDiameter);
+        properties._diameters.insert(properties._diameters.begin(), childSectionNextDiameter);
     }
 
     bool parse_neurite_section(int32_t parent_id, Token token) {
--- a/src/readers/morphologyASC.cpp
+++ b/src/readers/morphologyASC.cpp
@@ -150,6 +150,7 @@ class NeurolucidaParser
     /*
       Add the last point of parent section to the beginning of this section
       if not already present.
+      See https://github.com/BlueBrain/MorphIO/pull/221
 
       The diameter is taken from the child section next point as does NEURON.
       Here is the spec:
@@ -158,19 +159,19 @@ class NeurolucidaParser
       In term of diameters, the result should look like the right picture of:
       https://github.com/BlueBrain/NeuroM/issues/654#issuecomment-332864540
 
-      The idea is that these two structures should represent the same morphology:
+     The idea is that these two structures should represent the same morphology:
 
-      (3 -8 0 2)     and          (3 -8 0 2)
-      (3 -10 0 2)                 (3 -10 0 2)
-      (                           (
-        (0 -10 0 2)                 (3 -10 0 2)  <-- duplicate parent point
-        (-3 -10 0 2)                (0 -10 0 2)
-        |                           (-3 -10 0 2)
-        (6 -10 0 2)                 |
-        (9 -10 0 2)                 (3 -10 0 2)  <-- duplicate parent point
-      )                             (6 -10 0 2)
-                                    (9 -10 0 2)
-                                  )
+     (3 -8 0 5)     and          (3 -8 0 5)
+     (3 -10 0 5)                 (3 -10 0 5)
+     (                           (
+       (0 -10 0 2)                 (3 -10 0 2)  <-- duplicate parent point, note that the
+       (-3 -10 0 2)                (0 -10 0 2)      diameter is 2 and not 5
+       |                           (-3 -10 0 2)
+       (6 -10 0 2)                 |
+       (9 -10 0 2)                 (3 -10 0 2)  <-- duplicate parent point, note that the
+     )                             (6 -10 0 2)      diameter is 2 and not 5
+                                   (9 -10 0 2)
+                                 )
      */
     void insertLastPointParentSection(int32_t parentId,
                                       morphio::Property::PointLevel& properties,
--- a/tests/data/nested_single_children.asc
+++ b/tests/data/nested_single_children.asc
@@ -5,14 +5,15 @@
 
 ( (Color Blue)
   (Axon)
-  (0 0 0 1)
-  (0 0 1 1)
+  (0 0 0 8)
+  (0 0 1 7)
   (
-    (0 0 2 1)
+    (0 0 2 6)
     (
-      (0 0 3 1)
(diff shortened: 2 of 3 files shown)


wvangeit (approved):

src/readers/morphologyASC.cpp
```diff
@@ -152,32 +151,41 @@ class NeurolucidaParser
...
+      The diameter is taken from the child section next point as does NEURON.
+      Here is the spec:
+      https://bbpteam.epfl.ch/project/issues/browse/NSETM-1178?focusedCommentId=135030&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-135030
+
+      In term of diameters, the result should look like the right picture of:
+      https://github.com/BlueBrain/NeuroM/issues/654#issuecomment-332864540
+
+      The idea is that these two structures should represent the same morphology:
+
+      (3 -8 0 2)     and          (3 -8 0 2)
```
arnaudon, line 163: I would use non-constant diameters here to make the point clear for users nto reading this PR.

arnaudon (approved):

eleftherioszisis (approved):

src/readers/morphologyASC.cpp
wizmer: Very true !

mgeplf:
> When a single child section is merged with its parent section and the child section has a duplicate with a different diameter value, the diameter of the parent section is kept.

Is this what NEURON does?

wizmer:
As far as I know, NEURON does not merge sections, no ?

mgeplf:
> As far as I know, NEURON does not merge sections, no ?

I wasn't sure, it that description is under the banner of 'The following specification has been created to exactly recreate NEURON's behavior:', so it seemed counter-intuitive.

src/readers/morphologyASC.cpp
```diff
@@ -151,33 +150,43 @@ class NeurolucidaParser
...
-    void insertLastPointParentSection(int32_t parentId, morphio::Property::PointLevel& properties) {
+    void insertLastPointParentSection(int32_t parentId,
+                                      morphio::Property::PointLevel& properties,
+                                      std::vector<morphio::floatType>& diameters) {
         if (parentId < 0)  // Discard root sections
             return;
         auto parent = nb_.section(static_cast<unsigned int>(parentId));
         auto lastParentPoint = parent->points()[parent->points().size() - 1];
-        auto lastParentDiameter = parent->diameters()[parent->diameters().size() - 1];
+        auto childSectionNextDiameter = diameters[0];
```
mgeplf, line 183: Out of curiousity, where is diameters checked to have size >= 1?

wizmer: `insertLastPointParentSection` is called by `_create_soma_or_section` which is called by `parse_neurite_section` with the following code which discards empty sections:

```cpp
if (!points.empty()) {
    _create_soma_or_section(token, parent_id, points, diameters);
}
```

mgeplf: cool, thanks; and diameters & points is always mirrored?

mgeplf:
Adding a reviewer from HPC, so they can judge the impact of this change.

src/readers/morphologyASC.cpp
wizmer: Yes. And I'm confident that this would be caught by the unit tests otherwise. https://github.com/BlueBrain/MorphIO/blob/master/tests/test_2_neurolucida.py is full of edge cases.

wizmer:
That's gonna break so many tests everywhere lol

matz-e:
> Adding a reviewer from HPC, so they can judge the impact of this change.

Yeah, but wrong one. I'm clueless as to what NEURON™ does, I just took HDF5 as truth so far :)

I'm adding @pramodk 



> That's gonna break so many tests everywhere lol

Like changing the default behavior to merge unifurcations? :zany_face: 

Will there be a new version à la semantic thingies that clearly indicates that behavior has been changed? Will this affect morphologies that have been de-unifurcated?

wizmer:
> Like changing the default behavior to merge unifurcations? 

Oh, no, no. This change is much bigger than merging the unifurcations. It will affect almost every morphologie (except the ones that explicitely have duplicate points). It will change the diameter of all first points after each bifurcation which has no explicit duplicate.

> Will there be a new version à la semantic thingies that clearly indicates that behavior has been changed?

I don't exactly know what do you mean here. MorphIO attempts to read an unspecified proprietary format, the behavior might change whenever our comprehension of the format improves. Since I'm not sure bumping the major version every time is a solution I was planning to bump only the minor version number. Moreover, it's not an API breaking change, more like a bug fix (such as merging the unifurcation btw). Also, similarly when `scipy` changes the return value of a function, they also simply bump the minor version number, not the major. I'm open to any suggestion though, I understand this can be problematic.

matz-e:
Thank you for clarifying! The pipelines I'm working with are still mostly HDF5, so I don't consider myself affected (right? this is only for ascii). But thank you for the heads up!

From HPC's side, I think it would still be good if @pramodk and @alexsavulescu are aware of this, I can ping them separately on Slack, too

pramodk:
> When a single child section is merged with its parent section and the child
section has a duplicate with a different diameter value, the diameter of the
parent section is kept.

I don't have overall picture but want to mention following items here:
* there were couple of issues in handling SWC morphologies in NEURON https://github.com/neuronsimulator/nrn/pull/907
* significant changes/bug fixes were done recently in https://github.com/neuronsimulator/nrn/issues/750
* there is now new [python test here](https://github.com/neuronsimulator/nrn/pull/907/files#diff-5455098818adb7e1d3469912449d70790bdf2910a8f09ff390dd3628216a202b) which one can take a look at

alexsavulescu:
@wizmer I'm wondering if having duplicate points is linked somehow to the NEUROLUCIDA having unifurcations(single child) in the first place. AFAIK NEURON is not merging those sections. 

> The following specification has been created to exactly recreate NEURON's behavior:
> 
> * When a duplicate point is **not** in the file (ie. a section does not start at the
>   coordinate of its parent section last point), the duplicate is added. The
>   diameter of the added point is the diameter of the point that comes right after
>   it.
> * When a duplicate is already in the file, its coordinates and diameter value
>   are used as they are.
> * When a single child section is merged with its parent section and the child
>   section has a duplicate with a different diameter value, the diameter of the
>   **parent** section is kept.

@nrnhines could we please have your thoughts on these points?

nrnhines:
> When a single child section is merged with its parent section and the child
section has a duplicate with a different diameter value, the diameter of the
parent section is kept.

Whatever Import3d is doing, this does not seem right to me. I would expect the duplicate to remain with a different diameter value (discontinuous diameter change in the single child case). But maybe this is what that means. ie. the last point of the old parent section has the diameter specified by that point, and the diameter of the next point (the first point of the old child section) which  is 0 distance from the parent  has the diameter of the first point of the old child section and the new merged section continues as usual with the remaining child points.

wizmer:
I asked @asanin-epfl to help me investigate NEURON's behavior in the single point use case with this little snippet:
```
print(neuron.__version__)
h.load_file('single-child-template.hoc')
complex_cell = h.SingleChild(0, '.', 'single-child-morph2.asc')
print(complex_cell)
for sec in complex_cell.all:
    print(sec, 'len', sec.L)
    for seg in sec:
        print('diam', seg.diam)
```
We ran it with the following morphs:
## morph 1:
```lisp
("CellBody"
 (Color Red)
 (CellBody)
 (0 0 0 2)
 )
 ((Dendrite)
  (0 0 0 2)
  (0 5 0 2)
  (
   (0 5 0 3)
   (-5 5 0 3)
   )
  )
```

### result 

morph 2

wizmer:
I asked @asanin-epfl to help me investigate NEURON's behavior in the single point use case. He wrote this little snippet:
```
print(neuron.__version__)
h.load_file('single-child-template.hoc')
complex_cell = h.SingleChild(0, '.', 'single-child-morph2.asc')
print(complex_cell)
for sec in complex_cell.all:
    print(sec, 'len', sec.L)
    for seg in sec:
        print('diam', seg.diam)
```
We ran it with the following morphs:
## morph 1:
```lisp
("CellBody"
 (Color Red)
 (CellBody)
 (0 0 0 2)
 )
 ((Dendrite)
  (0 0 0 2)
  (0 5 0 2)
  (
   (0 5 0 3)
   (-5 5 0 3)
   )
  )
```

### result
```
SingleChild[0]
SingleChild[0].dend[0] len 5.0
diam 2.0
SingleChild[0].dend[1] len 5.0
diam 3.0
```

## morph 2

```lisp
("CellBody"
 (Color Red)
 (CellBody)
 (0 0 0 2)
 )
 ((Dendrite)
  (0 0 0 2)
  (0 5 0 2)
  (
   (0 5 0 2)  ; <- this 3->2 is the only change
   (-5 5 0 3)
   )
  )
```

### result

```
SingleChild[0]
SingleChild[0].dend[0] len 5.0
diam 2.0
SingleChild[0].dend[1] len 5.0
diam 2.5
```

## Interpretation

So as expected, NEURON does not merge single child sections. So in that case the behavior is straightforward, the diameter of the single child is the mean of both of its diameter points.

Actually, I'm not a NEURON specialist but I would be curious to know if these two dendrites are electrically equivalent:

morph 2 (same as before):

```lisp
((Color Red)
 (CellBody)
 (0 0 0 2)
 )

 ((Dendrite)
  (0 0 0 2)
  (0 5 0 2)
  (
   (0 5 0 2)
   (-5 5 0 3)
   )
  )
```

morph 3:

```lisp
((Color Red)
 (CellBody)
 (0 0 0 2)
 )

 ((Dendrite)
  ( 0 0 0 2)
  ( 0 5 0 2)
  (-5 5 0 3)
  )
```


However, in MorphIO it's a different story. The following paragraph is not about NEURON, just MorphIO. In MorphIO, we want to remove them (I think?) because:

1. They do not biologically represent anything. When looking at the microscope, a single child section does not exist. The piece of a dendrites that follows the previous piece of a dendrite is just the continuation of the same section.

2. Single child sections introduce computational bugs. It's important to be able to assume that a section has at least 2 children. Otherwise, it breaks many things. For example, in NeuroM all functions that work with bifurcation would no longer work. I don't know if unifurcation could have effects on Touch Detector also.


In MorphIO merging single child section is usually straightforward, except when the diameter differs before and after the unifurcation.

Should morph 3 (same as before):

```lisp
((Color Red)
 (CellBody)
 (0 0 0 2)
 )

 ((Dendrite)
  (0 0 0 2)
  (0 5 0 2) ; parent diameter: 2
  (
   (0 5 0 2) ; child diameter: 3
   (-5 5 0 3)
   )
  )
```

become

- morph 4:

```lisp
((Color Red)
 (CellBody)
 (0 0 0 2)
 )

 ((Dendrite)
  (0 0 0 2)
  (0 5 0 2) ; kept the parent diameter
  (-5 5 0 3)
  )
```

- morph 5:

```lisp
((Color Red)
 (CellBody)
 (0 0 0 2)
 )

 ((Dendrite)
  ( 0 0 0 2)
  ( 0 5 0 3) ; kept the child diameter
  (-5 5 0 3)
  )
```
- or even morph 6

```lisp
((Color Red)
 (CellBody)
 (0 0 0 2)
 )

 ((Dendrite)
  ( 0 0 0 2)
  ( 0 5 0 2.5) ; mean of the diameters
  (-5 5 0 3)
  )
```

> I would expect the duplicate to remain with a different diameter value (discontinuous diameter change in the single child case).

In fact, the only way to have discontinuous diameter changes at a given point, is when you have a bifurcation (regardless on the number of children).
For a point which is not at the beginning or the end of a section, it is not possible to have a different diameter on the left (at an epsilon distance) and on the right (at an epsilon distance) of the point. And that's why we need to make an assumption on the value  if we decide to merge these points.

The reason why, here we can take the liberty to have a behavior which is not the same as NEURON (merging vs not merging unifurcation) is that normally, within the Blue Brain, all morphologies must have been read and then re-written to disk using morphio. So if MorphIO removes all unifurcations, then we don't really need to know how NEURON deals with them, because it won't have to deal with any. So inconsistencies on this matter on not important.

alexsavulescu:
@wizmer FYI, we will explore together with @nrnhines in the following months the possibility of using MorphIO in NEURON for morphology loading. Would be awesome to have the possibility of doing exactly what NEURON does by default ..

wizmer:
I'm about to merge this. Any objections ?

PR #227, opened 2021-02-04
Stop supporting h5v2

wizmer:
If a h5v2 file is detected at loading time, the error message will point the user to the documentation for a way to convert it to h5v1.

https://github.com/BlueBrain/MorphIO#h5v2


tomdele:
Just for me to understand the format versioning. We drop the v2 but keep the v1 ?

wizmer:
Yes, v2 is a format that may have been used at some point, but whose use was stopped before I joined.
It was supposed to store multiple versions of the same morphology (raw morph, morph unravelled but not repaired, morph unraveled and repaired).  
The official latest supported version of h5 is h5v1.2

H5v2 is documented here:
https://bbpteam.epfl.ch/documentation/projects/Morphology%20Documentation/latest/h5v2.html
I will make a PR on this repo as well.

tomdele:
OK thanks

include/morphio/properties.h
```diff
@@ -151,6 +151,8 @@ struct Marker {
 };
 
 struct CellLevel {
+    CellLevel()
+        : _version(MORPHOLOGY_VERSION_UNDEFINED) {}
     morphio::CellFamily _cellFamily;
     SomaType _somaType;
     MorphologyVersion _version;
```
tomdele, line 158: minor : _version being part of the initialization list now you can move it to the top of the member declaration.

src/readers/morphologyHDF5.cpp
```diff
@@ -140,58 +98,73 @@ void MorphologyHDF5::_resolveV1() {
...
-        const auto familyAttr = metadata.getAttribute(_a_family);
-        familyAttr.read(family);
-    } catch (const HighFive::GroupException&) {
-        return false;
-    } catch (const HighFive::Exception& e) {
-        // All other exceptions are not expected because if the metadata
-        // group exits it must contain at least the version, and for
-        // version 1.1 it must contain the family.
-        throw morphio::RawDataError("Error reading morphology " + _uri + "  metadata." + e.what());
+        if (version[0] == 1 && version[1] >= 1) {
```
tomdele, line 119: Shouldn't we add an upper limit for the minor version number ?
Only 1.1 and 1.2 are supported and we want to throw if a 1.3 arrives one day, no ?

src/readers/morphologyHDF5.cpp
```diff
@@ -140,58 +98,73 @@ void MorphologyHDF5::_resolveV1() {
...
+            uint32_t family;
+            const auto familyAttr = metadata.getAttribute(_a_family);
+            familyAttr.read(family);
+            _properties._cellLevel._version = MORPHOLOGY_VERSION_H5_1_1;
```
tomdele, line 123: it can be 1.2 also here no ?

wizmer: Yes, you are totally correct.

src/readers/morphologyHDF5.cpp
wizmer: Yes, I was wondering actually. Indeed, it is probably safer to raise.

tomdele (approved):

tomdele:
except the failing test it looks good. :D

wizmer:
> except the failing test it looks good. :D

Yet another win for `git rebase`

wizmer:
Thank you !

PR #229, opened 2021-02-05
Improve marker handling

wizmer:
This commit adds support to markers suffixed with an arbitrary integer.

For example `Flower` is a valid marker but `Flower4` or `Flower4444` are now
also valid.

The list of valid marker tags has been updated based on the current morphology
repository that we have.

:information_source: In our morphology repository, only `Circle` and `Flower` have been found with suffixes.

## Marker specification
Here are the regexes of supported markers:

- Dot[0-9]*
- Plus[0-9]*
- Cross[0-9]*
- Splat[0-9]*
- Flower[0-9]*
- Circle[0-9]*
- Flower[0-9]*
- TriStar[0-9]*
- OpenStar[0-9]*
- Asterisk[0-9]*
- SnowFlake[0-9]*
- OpenCircle[0-9]*
- ShadedStar[0-9]*
- FilledStar[0-9]*
- TexacoStar[0-9]*
- MoneyGreen[0-9]*
- DarkYellow[0-9]*
- OpenSquare[0-9]*
- OpenDiamond[0-9]*
- CircleArrow[0-9]*
- CircleCross[0-9]*
- OpenQuadStar[0-9]*
- DoubleCircle[0-9]*
- FilledSquare[0-9]*
- MalteseCross[0-9]*
- FilledCircle[0-9]*
- FilledDiamond[0-9]*
- FilledQuadStar[0-9]*
- OpenUpTriangle[0-9]*
- FilledUpTriangle[0-9]*
- OpenDownTriangle[0-9]*
- FilledDownTriangle[0-9]*

## Code snippet

Here is the code I used to generate the list of markers. I then had to manually
filter and check the result to obtain the final list of markers.

```python
import json
from itertools import chain

from morph_tool.utils import iter_morphology_files
from tqdm import tqdm

def create_token_list():
    files = list(iter_morphology_files('/home/bcoste/workspace/MorphologyRepository/',
                                       recursive=True))

    def is_not_number(token):
        try:
            float(token)
            return False
        except ValueError:
            return True

    def parse_line(line):
        tokens = line.split(';')[0].replace('(', ' ').replace(')', ' ').split()
        return filter(is_not_number, tokens)
    token = set()
    for f in tqdm(files):
        try:
            new_set = set(chain.from_iterable(parse_line(line) for line in f.open().readlines()))
            token = token | new_set
        except:
            print(f'failed parsing {f}')

    with open('bla.json', mode='w') as f:
        json.dump(list(token), f)

def read_token_list():
    with open('bla.json') as f:
        data = json.load(f)

    filtered = {'Settings\\user\\Desktop\\tkb02\\060206\\'}

    data = filter(lambda token: not token.startswith('0x'), data)
    data = filter(lambda token: token[0].isupper(), data)
    data = filter(lambda token: all(word not in token for word in filtered), data)
    data = sorted(data, key=lambda token: len(token))
    data = list(data)

    from pprint import pprint
    pprint(data)
```


arnaudon:
Just trying the previous version on some morphs:
with this:
```
("contour outline"                                            
  (Color White)                                               
  (Closed)                                                    
  (GUID "5B0C7BBAF4494E41A134383FE8D8104C")                                
  (MBFObjectType 5)                                           
  (FillDensity 0)                                             
  (Resolution 2.822065)                                                     
  ( 4815.65  5181.83     3.00     0.10 S1)  ;  1, 1           
  ( 4897.49  5145.14     3.00     0.10 S1)  ;  1, 2           
  ( 5013.19  5108.46     3.00     0.10 S1)  ;  1, 3     
```
I see a good `morph.markers`, with label = contour outline. 

If I load the errors from neurom, for example:
```
                                                              
(Circle2   ; MUK_ANNOTATION                                   
    (Color Green)   ; MUK_ANNOTATION                          
    (Name "zjump")   ; MUK_ANNOTATION                         
    (     58.95     186.75     158.22 0.50)   ; MUK_ANNOTATION
    (     61.16     225.10     190.04 0.50)   ; MUK_ANNOTATION
    (     58.21     252.63     218.96 0.50)   ; MUK_ANNOTATION
    (     39.26     292.67     260.24 0.50)   ; MUK_ANNOTATION
    (     39.26     292.67     260.24 0.50)   ; MUK_ANNOTATION
    (    -15.41     360.09     313.92 0.50)   ; MUK_ANNOTATION
```
I get an empty `marker.label=''`, but we would like to have ``marker.label='zjump`, no?

wizmer:
Hum, yes, but that can probably be considered as an extra feature because that would require to add the parsing of (Name) tags, which we don't do yet. But yes.

arnaudon:
Instead, can we make our marker the same as the other one? So the label goes through the same?

arnaudon:
@wizmer , sorry, don't bother with that, this is for ying only. I'll modify my PR to save the markers in another file for plotting.

arnaudon (approved):

tomdele:
Do I understand correctly : 
do we have some plotting properties included inside the data format ?

wizmer:
In Neurolucida, when reconstructor are looking at the microscope, they can click on the neuron points to reconstruct the neuron, but they can also click on anything else and save it in the same file. This is what we call markers. There are several use of the markers: 
- experimentalists use them to save additional information (for example the location of the pia)
- We use them at BBP to highlight some specific location in the morphology where NeuroM found an anomaly. Experimentalists then open the marked file in the NeuroLucida and the marker location will be highlighted so they can have a look and fix the detected anomaly.

wizmer:
Markers are just a list of coordinates, a "shape" (how the marker will look on Neurolucida. Ex: a star, a circle, a filled circle...), and a label.

tomdele:
That's a bit weird for a science format to use a random marker shape instead of a "event" (like "info", "warning", "danger" or even "pia") that is read and then represented by the different viewers. It looks like a svg marker... 
But ok I got it.

tomdele (approved):

arnaudon:
I'm drafting an internal API to handle these ;) I'll add you guys to review, so we can discuss further!

arnaudon:
Here it is: https://github.com/BlueBrain/morph-tool/pull/47

mgeplf:
> I then had to manually filter and check the result to obtain the final list of markers.

Could you give the list of which ones you found?

I still think it's worth seeing what neurolucida does, especially with edge-cases like `Flower00003`, `Flower12341234`, etc.

Perhaps we can get someone to annotate a fake morphology w/ the min/max (ie: `Circle0`/`Circle9`) of each of the markers, so we don't accept things that don't exist.

wizmer:
The only markers I found that were ending with numbers were:
- Circle1
- Circle2
- Circle3
- Circle6
- Flower2

mgeplf:
Cool, thanks.  I guess we have no idea if 'Circle' (w/ no number) is valid?  I guess we'll just have to be liberal until someone provides a set of fake annotations w/ all the markers.

wizmer:
Ah, no. I just listed the ones who had numbers but `Circle` and `Flower` also exists without any suffix.

PR #231, opened 2021-02-05
Refactor versioning

wizmer:
No longer use an enum as the return type of `Morphology::version()` but a tuple <string, int, int>. The three fields are:

- file format extension. One of "h5", "asc", "swc"
- major version for the given format
- minor version for the given format

So far the existing values are:
- ("swc", 1, 0)
- ("asc", 1, 0)
- ("h5", 1, 0)
- ("h5", 1, 1)
- ("h5", 1, 2)

The advantage is that versions can now be ordered within a given file format.

* Fix the metadata version field in H5. It is now 1.2 instead of 1.1 as per the spec:  https://bbpteam.epfl.ch/documentation/projects/Morphology%20Documentation/latest/h5v1.html

### DIFF
--- a/.github/workflows/publish-sdist-wheels.yml
+++ b/.github/workflows/publish-sdist-wheels.yml
@@ -1,9 +1,10 @@
 name: Publish sdist and wheels macos-manylinux
 
 on:
-  push:
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+  pull_request:
+  # push:
+  #   tags:
+  #     - 'v[0-9]+.[0-9]+.[0-9]+'
 
 env:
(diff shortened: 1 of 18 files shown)


wizmer:
@matz-e Do you know if this attribute is used in the HPC stack ? If not, I would like to propose this PR. Otherwise I can keep the old `version` and create a new method (`version_tuple` or something).

matz-e (approved):
@wizmer From my side, this is OK. I can't really think of a use case where we query the version attribute and act on that.

OTOH, this seems like a behavior change, are we going to at least bump the minor version for this?

wizmer:
Thanks Matthias. Yes, I will bump the minor version.

PR #233, opened 2021-02-10
Allow markers to have no diameters

wizmer:
Markers may have only `(X Y Z)` specified instead of the more common `(X Y Z D)`. In this case, diameters are set to 0.

Also:
Strings in the middle of the list of points are now skipped.

ℹ️ The example file added in this commit is based on C280999A-I4.asc
It illustrates both use cases.

### DIFF
--- a/src/readers/morphologyASC.cpp
+++ b/src/readers/morphologyASC.cpp
@@ -76,23 +76,30 @@ class NeurolucidaParser
     }
 
   private:
-    std::tuple<Point, floatType> parse_point(NeurolucidaLexer& lex) {
+    std::tuple<Point, floatType> parse_point(NeurolucidaLexer& lex, bool is_marker) {
         lex.expect(Token::LPAREN, "Point should start in LPAREN");
-        std::array<morphio::floatType, 4> point{};  // X,Y,Z,R
-        for (auto& p : point) {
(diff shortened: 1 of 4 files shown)


arnaudon (approved):

src/readers/morphologyASC.cpp
```diff
@@ -76,23 +76,30 @@ class NeurolucidaParser
...
 #else
-                p = std::stof(lex.consume()->str());
+                point[i] = std::stof(lex.consume()->str());
 #endif
             } catch (const std::invalid_argument&) {
                 throw RawDataError(err_.ERROR_PARSING_POINT(lex.line_num(), lex.current()->str()));
             }
+
+            // Markers can have an s-exp (X Y Z) without diameter
+            if (is_marker && i == 2 && (lex_.peek()->str() == ")")) {
```
tomdele, line 94: haha ok. There is no other way than `")"` to check you reached the end ? Or something like a lex_.size() == 3 ?

wizmer: It's a parser so you don't know what's ahead of you, you can simply "peek" at the next token :)
Otherwise that would be too easy.

tomdele: pouet

tomdele (approved):

wizmer:
Thanks pouet.

Issue #161, opened 2020-02-14
Lifetime issues with mut.Morphology constructing sections with raw pointers to itself


mgeplf:
Constructs like this:
`Section(this, _counter, section_)` (where `this` is mut::Morphology` - [from](https://github.com/BlueBrain/MorphIO/blob/master/src/mut/morphology.cpp#L89))

This means that a raw pointer is stored [here](https://github.com/BlueBrain/MorphIO/blob/master/include/morphio/mut/section.h#L109) that has a different lifetime than `Morphology`.

This causes a use after free if the Morphology is dtor'd before the sections that have been created are removed, for example:
```
from morphio.mut import Morphology
m = Morphology('simple.swc')
s = m.root_sections[0]
del m # ~mut.Morphology() called
print(s.children) # use after free
```

Note, this is different than #29

wizmer:
I have been thinking about this for a while. For me what seemed weird is that by design, a section is not supposed to be used after the morphology has been deleted. This is the analogue of attempting to use a node after you have deleted a graph. 

One proposal I have is to add a method '_raise_if_not_alive()' call at the beginning of each function so that it raises an actual error when the user attempts to do the manipulation you wrote. 

Would that be good?

mgeplf:
> For me what seemed weird is that by design, a section is not supposed to be used after the morphology has been deleted.

That's the simplified test case to aide in debugging.

The issue is that one has to know this, and work around it (hence the 'Hack for morphio lifetime issue`.)  Since python is a GC'd language, the expected behavior is that it should stay alive as long as you have a reference to it.

Currently this is causing corruption that is silent, hard to track down, and can cause problems later.  `_raise_if_not_alive` would be a start, so at least it wouldn't be silent, but it would also need to not allow corruption.  It also seems like you may have the same lifetime issues.

Ping @tomdele so he can be up to speed on this.

tomdele:
To me this problem is kinda similar to this one I had in snap:
https://github.com/BlueBrain/snap/pull/42

I spent hours on this and could not find anything usable. The del from python is a much more complex beast than the c++ dtors (even if more handy). 
The del will remove your variable name from the global scope but the actual `__del__()` is called only when the pointer counters reached 0 for all objects. 

So this is not specific to MorphIO and easily reproducible in python only:
```
class B:
    def __init__(self, a_object, index):
        self.a = a_object
        self.index = index

    def get_next(self):
        return self.a.get_at(self.index + 1)


class A:
[...]

PR #234, opened 2021-02-15
Fix lifetime issue of mutable morphs #161

wizmer:
mut::Morphology dtor no longer calls mut::Section dtors. Instead it unlinks them
so they become flagged as unusable. Any attempt to use them will result in an
RuntimeError exception.

The python garbage collector becomes the guarantor of the sections deletion.

### DIFF
--- a/binds/python/bind_mutable.cpp
+++ b/binds/python/bind_mutable.cpp
@@ -21,21 +21,34 @@ void bind_mutable_module(py::module& m) {
     using namespace py::literals;
 
 
-    py::class_<morphio::mut::Morphology>(m, "Morphology")
+    py::class_<morphio::mut::Morphology, std::shared_ptr<morphio::mut::Morphology>>(m, "Morphology")
         .def(py::init<>())
-        .def(py::init<const std::string&, unsigned int>(),
+        .def(py::init([](const std::string& filename, unsigned int options){
+            auto ptr = std::make_shared<morphio::mut::Morphology>();
+            ptr->init(filename, options);
+            return ptr;
+        }),
              "filename"_a,
              "options"_a = morphio::enums::Option::NO_MODIFIER)
-        .def(py::init<const morphio::Morphology&, unsigned int>(),
-             "morphology"_a,
-             "options"_a = morphio::enums::Option::NO_MODIFIER)
-        .def(py::init<const morphio::mut::Morphology&, unsigned int>(),
+        .def(py::init([](const morphio::Morphology& morph, unsigned int options) {
+            auto ptr = std::make_shared<morphio::mut::Morphology>();
+            ptr->init(morph, options);
+            return ptr;
+        }),
+            "morphology"_a,
+            "options"_a = morphio::enums::Option::NO_MODIFIER)
+        .def(py::init([](const morphio::mut::Morphology& morph, unsigned int options) {
+            auto ptr = std::make_shared<morphio::mut::Morphology>();
+            ptr->init(morph, options);
+            return ptr;
+        }),
              "morphology"_a,
              "options"_a = morphio::enums::Option::NO_MODIFIER)
         .def(py::init([](py::object arg, unsigned int options) {
-                 return std::unique_ptr<morphio::mut::Morphology>(
-                     new morphio::mut::Morphology(py::str(arg), options));
-             }),
+            auto ptr = std::make_shared<morphio::mut::Morphology>();
+            ptr->init(py::str(arg), options);
+            return ptr;
+        }),
              "filename"_a,
              "options"_a = morphio::enums::Option::NO_MODIFIER,
              "Additional Ctor that accepts as filename any python "
@@ -178,16 +191,17 @@ void bind_mutable_module(py::module& m) {
              "mutable_section"_a,
              "recursive"_a = false);
 
-    py::class_<morphio::mut::GlialCell, morphio::mut::Morphology>(m, "GlialCell")
+    py::class_<morphio::mut::GlialCell, morphio::mut::Morphology,
+               std::shared_ptr<morphio::mut::GlialCell>>(m, "GlialCell")
         .def(py::init<>())
-        .def(py::init<const std::string&>())
-        .def(py::init([](py::object arg) {
-                 return std::unique_ptr<morphio::mut::GlialCell>(
-                     new morphio::mut::GlialCell(py::str(arg)));
-             }),
-             "filename"_a,
-             "Additional Ctor that accepts as filename any python "
-             "object that implements __repr__ or __str__");
+        .def(py::init([](py::object arg, unsigned int options) {
+            auto ptr = std::make_shared<morphio::mut::GlialCell>();
+            ptr->init(py::str(arg), options);
+            return ptr;
+        }),
+            "filename"_a,
+            "options"_a = morphio::enums::Option::NO_MODIFIER,
+            "Ctor from a existing file");
 
 
     py::class_<morphio::mut::Mitochondria>(m, "Mitochondria")
--- a/binds/python/bind_mutable.cpp
+++ b/binds/python/bind_mutable.cpp
@@ -191,17 +191,18 @@ void bind_mutable_module(py::module& m) {
              "mutable_section"_a,
              "recursive"_a = false);
 
-    py::class_<morphio::mut::GlialCell, morphio::mut::Morphology,
+    py::class_<morphio::mut::GlialCell,
+               morphio::mut::Morphology,
                std::shared_ptr<morphio::mut::GlialCell>>(m, "GlialCell")
         .def(py::init<>())
         .def(py::init([](py::object arg, unsigned int options) {
-            auto ptr = std::make_shared<morphio::mut::GlialCell>();
-            ptr->init(py::str(arg), options);
-            return ptr;
-        }),
-            "filename"_a,
-            "options"_a = morphio::enums::Option::NO_MODIFIER,
-            "Ctor from a existing file");
+                 auto ptr = std::make_shared<morphio::mut::GlialCell>();
+                 ptr->init(py::str(arg), options);
+                 return ptr;
+             }),
+             "filename"_a,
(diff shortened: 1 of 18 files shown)


wizmer:
@mgeplf This is an attempt at fixing the issue. The main drawback I see is that this also makes the C++ users directly responsible for deleting the mutable sections. However, right now it is not really an issue since there are no mutable C++ API users.
This is just a draft, let me know what do you think and if you think it's worth continuing in this direction (updating the README and such)

mgeplf (commented):
I'm a bit concerned about this method, because I think it's a band-aid fix - there are a lot of places that `_raiseIfLink` need to be added.  Are you sure that you're considering all the cases on how sections can be moved around?

> The main drawback I see is that this also makes the C++ users directly responsible for deleting the mutable sections

Yeah, that's also true; it makes the ownership hard to reason about, as well.

I think it would be worth getting @tomdele's opinion since he'll be maintaining this.

src/mut/section.cpp
```diff
@@ -13,6 +13,13 @@ static inline bool _emptySection(const std::shared_ptr<Section>& section) {
     return section->points().empty();
 }
 
+void Section::_raiseIfUnlinked() const {
+    if (!_morphology)
+        throw std::runtime_error(
+            "This morphology can no longer be used because the Morphology "
```
tomdele, line 19: It is this:  section can no longer be used ... no ?

tomdele:
I put a comment here : 
https://github.com/BlueBrain/MorphIO/issues/161#issuecomment-780583576

binds/python/bind_mutable.cpp
```diff
@@ -21,21 +21,34 @@ void bind_mutable_module(py::module& m) {
...
 
-    py::class_<morphio::mut::Morphology>(m, "Morphology")
+    py::class_<morphio::mut::Morphology, std::shared_ptr<morphio::mut::Morphology>>(m, "Morphology")
         .def(py::init<>())
-        .def(py::init<const std::string&, unsigned int>(),
+        .def(py::init([](const std::string& filename, unsigned int options){
+            auto ptr = std::make_shared<morphio::mut::Morphology>();
+            ptr->init(filename, options);
+            return ptr;
+        }),
```
tomdele, line 30: Actually I was thinking about that yesterday and this will actually break inheritance, no? 
```
>>> class MyMorpho(Morphology):
>>>    def foo():
>>>          ...

>>> a = MyMorpho(file)
>>> type(a)
Morphology
>>> 
```

wizmer: Good question, you scared me but I just tried on ipython and it seems to work.
```
In [1]: class M(mio.mut.Morphology):
   ...:     pass
   ...:

In [2]: a = M('tests/data/simple.swc')

In [3]: a
Out[3]: <__main__.M at 0x10fc4c4d0>

In [4]: type(a)
Out[4]: __main__.M
```

tomdele: Cool I did not try yet !
I still need to understand exactly why but that s cool !

wizmer:
The PR is ready for review. I can't currently fix the CI since `./ci/check_clang_format.sh` currently gives
```
ERROR: Could not find a version that satisfies the requirement clang-format==9.0.0
```
on my mac but I'll fix this tomorrow on my linux machine.

mgeplf:
> The PR is ready for review. I can't currently fix the CI since `./ci/check_clang_format.sh` currently gives

Can you explain the methodology a bit?

wizmer:
Preface: Since the immutable API is left untouched and is not part of the discussion, the variable and class names used in the following will refer to the ones in the mutable namespace.

## API change
As before, the morphology object is a container of `shared_ptr<Section>` however the big change is that `Section` objects no longer own a raw pointer `Morphology*` but a `weak_ptr<Morphology>`.
This weak pointer is only used when trying to retrieve topological information like `Section::parent` or `Section::children`. Before each use of this pointer, a check is made on its ref count. If the ref count is 0, an exception will be thrown. The take away point is that if the parent morphology is deleted, a section can still be used but attempts to access any topological information will result in an exception.

## Implementation change
Implementation wise, the big changes are:
- On the c++ side, a Morphology object must no longer be created on the stack, it must be created with `std::make_shared<Morphology>`. This is because in order to pass the morphology weak pointer to the `Section` ctor, `Morphology::shared_from_this()` is used and requires an existing shared pointer.
- A corollary is that the registration of the `Section` objects can no longer be done in the `Morphology` ctor because the shared pointer has not yet been created. Hence, on the c++ sides instantiating an `Morphology` object from a file on disk or another morphology must be done in 2 steps: an invocation of the ctor followed by an initialization:
```c++
auto morph = std::make_shared<Morphology>();
morph->init(filename)
```

ℹ️  On the python side, no changes are required since these two steps have been wrapped in the pybind ctor of the `Morphology` class. To be clear, the user can keep using the following like in the past:
```python
morph = Morphology(filename)
```

ℹ️  Since the class `GlialCell` derives from `Morphology` it is subject to the same changes (can no longer be created without `std::make_shared` and requires a 2-step initialization).

tomdele:
Sorry I forgot to do the review today. I will do it tomorrow.

tests/test_5_mut.py
```diff
@@ -466,25 +464,34 @@ def test_sanitize():
...
+
+def test_lifetime_issue():
+    m = Morphology(DATA_DIR / 'simple.swc')
+    s = m.root_sections[0]
+    del m # ~mut.Morphology() called
+    with assert_raises(RuntimeError) as obj:
+        s.children
+    assert_substring("This section (id: 0) can no longer be used in the context of the whole "
+                     "morphology because the Morphology object it belongs to has been deleted",
+                     str(obj.exception))
```
mgeplf, line 497: can you add a few more tests here; ie: moving sections from morph to morph, deleting the section, deleting the morphs, etc, so it's obvious what is allowed and what isn't?

mgeplf:
Thanks for the description, makes a huge difference for understanding.  I still have to review though, but I have good context.

src/mut/section.cpp
```diff
@@ -22,33 +32,38 @@ Section::Section(Morphology* morphology,
     , _id(id_)
     , _sectionType(type_) {}
 
-Section::Section(Morphology* morphology, unsigned int id_, const morphio::Section& section_)
+Section::Section(std::weak_ptr<Morphology> morphology,
```
tomdele, line 35: Don t we want to .lock() the morphology in the constructor as a sanity check ? to avoid producing a section from a count ref = 0 morpho ?
Not sure if this can happen tho.

tomdele:
I think I am good with the changes. 

I will do a second path on this just to be sure I am not missing something.

src/mut/section.cpp
tomdele: (` .lock()`  plus if (shared) of course.

tristan0x:
IMO, the C++ current memory model in MorphIO is fine and decorrelating the liveness of C++ Section and Morphology:
* does not intrinsically address #234 (despite it fix it) because IMO #234 is liveness problem **in Python** because C++ and Python memory models are different.
* introduce complexity in the API in the sense that C++ code using a Section must now ask itself if the associated Morphology instance is still here.

-----

```python
from morphio.mut import Morphology
m = Morphology('simple.swc')
s = m.root_sections[0]
del m # ~mut.Morphology() called
print(s.children) # use after free
```

The issue here is that Python should know that `s` depends on `m`. This is a Python reference counting issue to me. `s`  (the Python object) should have a reference to the Morphology **Python** object so that `del m` does not really destroy `m`.

mgeplf:
@tristan0x Thanks very much for looking at this; it's very helpful.

> IMO, the C++ current memory model in MorphIO is fine and decorrelating the liveness of C++ Section and Morphology:

To confirm, you mean what exists in currently in `master`, not what is in this PR?

> introduce complexity in the API in the sense that C++ code using a Section must now ask itself if the associated Morphology instance is still here.

This is the crux of it, in my opinion; I don't think this is just a mismatch of C++/python memory models.  The current `mut::Section` has a backpointer to its `Morphology` (https://github.com/BlueBrain/MorphIO/blob/master/include/morphio/mut/section.h#L109); if that `Morphology` gets deallocated, the `mut::Section` then has a dangling pointer, which has ended up in hard to debug problems.

On the C++ side, I could imagine the same issue being triggered by something like (excuse my bad code):
```
using Sections = std::vector<std::shared_ptr<morphio::mut::Section>>;
Sections get_first_sections(std::vector<std::string> paths)
{
  Sections ret;
  for(auto path : paths){
    auto s = morphio::mut::Morphology(path).sections();
    ret.push_back(s[0]);
  }
  return ret;
}

morphio::mut::Morphology make_franken_morph(){
   Sections secs = get_first_sections({"data/simple.swc", "data/simple.swc"});

   auto new_m = morphio::mut::Morphology("data/simple.swc");
   for(auto sec: secs){
     new_m.appendRootSection(sec, true);
   }
   return new_m;
}
```

In this case, it's only the `std::shared_ptr<mut::Sections>` that are passed between functions, and the morphs get dtor'd.  This causes a failure when `appendRootSection` tries to access the dangling `Section::morphology` pointer; this is a slightly contrived example, but I don't think it's out of the ordinary to expect `shared_ptr` wrapped object to act like a value type, and not require other objects to be kept alive - however, I'm not up to date on `C++`, so maybe it's expected that this would work?

> The issue here is that Python should know that s depends on m. This is a Python reference counting issue to me. s (the Python object) should have a reference to the Morphology Python object so that del m does not really destroy m.

That's true, and that would be my preferred fix; then the behavior is pythonic; however, it should also still be `C++`ithic, so that we aren't just targetting python.

mgeplf:
Thoughts on the above @wizmer / @tomdele / @tristan0x ?

Setting `mut::Section::_morphology` to `nullptr` in the `Morphology` dtor would, I think, make the errors to detect, along w/ adding the `throw`ing of exceptions where that pointer is used.

wizmer:
Replaced by #253

PR #236, opened 2021-02-24
Make pypi tarball compilable

### LINKED ISSUE #232: MorphIO tarball should be compilable
Looks like the same problem as:
https://github.com/BlueBrain/libsonata/issues/117

Hopefully the same fix applies.

wizmer:
Fix #232

### DIFF
--- a/.github/workflows/publish-sdist-wheels.yml
+++ b/.github/workflows/publish-sdist-wheels.yml
@@ -1,9 +1,10 @@
 name: Publish sdist and wheels macos-manylinux
 
 on:
-  push:
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+  pull_request:
+  # push:
+  #   tags:
+  #     - 'v[0-9]+.[0-9]+.[0-9]+'
 
 env:
   CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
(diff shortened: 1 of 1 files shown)


wizmer:
I've tested that it works by temporary changing the trigger to `pull_requests:`.
Here is the succesful build where the tarball was built and tested:
https://github.com/BlueBrain/MorphIO/runs/1972132938?check_suite_focus=true

.github/workflows/publish-sdist-wheels.yml
```diff
@@ -62,14 +62,23 @@ jobs:
...
           python-version: 3.6
       - name: Build a source tarball
         run:
           python setup.py sdist
+      - name: Test tarball
+        run: |
+          sudo apt install -y libhdf5-dev
+          pip install -r tests/requirement_tests.txt
+          pip install dist/*
+          nosetests -s -v -P tests
```
tomdele, line 81: Stupid question :
run from actions propagates the returned "successful or not termination" of the command line ?

wizmer: Yes. https://docs.github.com/en/actions/creating-actions/setting-exit-codes-for-actions#about-exit-codes

tomdele: Au top topper !

tomdele (approved):

Issue #235, opened 2021-02-19
Allow for sanitization bypass for legacy


arnaudon:
It may be useful to have a hidden option to not sanitize (at least unifurcation merge), for legacy mostly.
There are two branches doing that curently:
https://github.com/BlueBrain/MorphIO/tree/no-sanitize
https://github.com/BlueBrain/MorphIO/tree/unifurcation

this option should be very hidden, so nobody can do it easily ;)

mgeplf:
As evidenced by the branches that do it, it's something that's come up a few times.

I think we should decide what the behavior is going forward, because the auto-magic merging has bitten us a number of times.

My view is that when a morphology is loaded, it should represent what is on disk as faithfully as possible; so as to not confuse unnecessarily the consumers - for instance the changing of `section_ids` post-sanitization.  A user could _choose_ to sanitize post loading if they wanted to, but that could be a section function call or a loading option - not the default.

The benifit of this is that it is backwards compatible with how morphs have been written, compatible with the morphs that are out there, and doesn't put the onus on the consume to track the changes if they need to deal with legacy.

I would like to hear/get a consensus from other consumers: specifically the HPC team (@matz-e, and anyone you recommend) and the VIZ team (@NadirRoGue, and anyone you recommend)

On the write side, when a format is being written that cannot support certain features (ie: SWC can't handle unifurcations), I believe this should raise an error, so the user can choose what to do (ie: they have to run `morph.sanitize()` or `eliminate_unifurcations(morph)`.  With in the `morphology-release` process, this could be required - however, that only changes release going forward.

NadirRoGue:
I agree with @mgeplf . An option to choose whether branch merge should be applied or not would be quite useful. For visualization, we should be able to show what the actual morphology looks like. 

Also, as @chevtche can explain, branch merging would affect the results of other programs, such as LFP calculations on EMSim.

matz-e:
Same here, I second/third @mgeplf. @alexsavulescu and @pramodk could also give their opinions?

alexsavulescu:
All in favour on my side. Older discussion here: https://github.com/BlueBrain/MorphIO/issues/175
Please note that https://github.com/BlueBrain/MorphIO/tree/unifurcation can be considered as a dirty-fix, since I was having some issues with implementing the CMake option (don't remember which ones, I vaguely remember due to switching between mutable/immutable). 

Additionally there is no way as of now to know if a circuit has unifurcations or not before simulation time. 
[...]

PR #238, opened 2021-03-02
MorphIO no longer merge unifurcations

wizmer:
This is true for loading **and** writing.

To be precise, it still merges them when **writing in SWC format** but that's because we do
not have the choice.

Consequently this means the definition of a `section` has been changed to:

```
Section:
- A succession of points between two bifurcations if the file is in SWC format.
- A succession of points (not necessarily between two bifurcations) in an S-exp if the file is in ASC format.
- A succession of points (not necessarily between two bifurcations) corresponding to a block in an H5 format.
```

This is the outcome of the discussion that occured in #235

### DIFF
--- a/include/morphio/mut/morphology.h
+++ b/include/morphio/mut/morphology.h
@@ -213,6 +213,11 @@ class Morphology
     void removeUnifurcations();
     void removeUnifurcations(const morphio::readers::DebugInfo& debugInfo);
 
+    /**
+       Used before writing to SWC to check that there is no unifurcation
+     **/
+    bool _checkUnifurcations();
+
   public:
     friend class Section;
     friend void modifiers::nrn_order(morphio::mut::Morphology& morpho);
--- a/include/morphio/mut/morphology.h
+++ b/include/morphio/mut/morphology.h
@@ -216,7 +216,7 @@ class Morphology
     /**
        Used before writing to SWC to check that there is no unifurcation
      **/
-    bool _checkUnifurcations();
+    void _raiseIfUnifurcations();
 
   public:
     friend class Section;
--- a/include/morphio/mut/morphology.h
+++ b/include/morphio/mut/morphology.h
@@ -210,8 +210,8 @@ class Morphology
        Fixes the morphology single child sections and issues warnings
        if the section starts and ends are inconsistent
      **/
-    void sanitize();
-    void sanitize(const morphio::readers::DebugInfo& debugInfo);
+    void removeUnifurcations();
+    void removeUnifurcations(const morphio::readers::DebugInfo& debugInfo);
 
   public:
     friend class Section;
--- a/src/morphology.cpp
+++ b/src/morphology.cpp
@@ -34,7 +34,6 @@ Morphology::Morphology(const Property::Properties& properties, unsigned int opti
     // their respective loaders
     if (properties._cellLevel.fileFormat() == "h5") {
         mut::Morphology mutable_morph(*this);
-        mutable_morph.sanitize();
         if (options) {
             mutable_morph.applyModifiers(options);
         }
@@ -50,7 +49,6 @@ Morphology::Morphology(const std::string& source, unsigned int options)
     : Morphology(loadURI(source, options), options) {}
 
 Morphology::Morphology(mut::Morphology morphology) {
-    morphology.sanitize();
     _properties = std::make_shared<Property::Properties>(morphology.buildReadOnly());
     buildChildren(_properties);
 }
--- a/src/mut/morphology.cpp
+++ b/src/mut/morphology.cpp
@@ -206,11 +206,11 @@ void _appendProperties(Property::PointLevel& to, const Property::PointLevel& fro
         _appendVector(to._perimeters, from._perimeters, offset);
 }
 
-void Morphology::sanitize() {
-    sanitize(morphio::readers::DebugInfo());
+void Morphology::removeUnifurcations() {
+    removeUnifurcations(morphio::readers::DebugInfo());
 }
 
-void Morphology::sanitize(const morphio::readers::DebugInfo& debugInfo) {
+void Morphology::removeUnifurcations(const morphio::readers::DebugInfo& debugInfo) {
     morphio::readers::ErrorMessages err(debugInfo._filename);
 
     auto it = depth_begin();
@@ -368,7 +368,7 @@ void Morphology::write(const std::string& filename) {
         // the concept of section does not have any meaning for the file format.
         // And due to how the swc writer() works, the file needs to be cleaned
         // or the parent ID will be messed up
-        clean.sanitize();
+        clean.removeUnifurcations();
         writer::swc(clean, filename);
     } else
         throw UnknownFileType(_err.ERROR_WRONG_EXTENSION(filename));
--- a/src/mut/morphology.cpp
+++ b/src/mut/morphology.cpp
@@ -364,6 +364,10 @@ void Morphology::write(const std::string& filename) {
         writer::asc(*this, filename);
     else if (extension == ".swc") {
         morphio::mut::Morphology clean(*this);
+        // Due to the structure based on IDs and parent IDs of SWC
+        // the concept of section does not have any meaning for the file format.
+        // And due to how the swc writer() works, the file needs to be cleaned
+        // or the parent ID will be messed up
         clean.sanitize();
         writer::swc(clean, filename);
     } else
--- a/src/mut/morphology.cpp
+++ b/src/mut/morphology.cpp
@@ -260,6 +260,23 @@ void Morphology::removeUnifurcations(const morphio::readers::DebugInfo& debugInf
     }
 }
 
+bool Morphology::_checkUnifurcations() {
+    for (auto it = depth_begin(); it != depth_end(); ++it) {
+        std::shared_ptr<Section> section_ = *it;
+        if (section_->isRoot())
+            continue;
+
+        unsigned int parentId = section_->parent()->id();
+
+        auto parent = section_->parent();
+        bool isUnifurcation = parent->children().size() == 1;
+
+        if (isUnifurcation) {
+            throw WriterError(readers::ErrorMessages().ERROR_ONLY_CHILD_SWC_WRITER(parentId));
+        }
+    }
+}
+
 Property::Properties Morphology::buildReadOnly() const {
     using std::setw;
     int sectionIdOnDisk = 0;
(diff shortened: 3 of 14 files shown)


tests/test_2_neurolucida.py
```diff
@@ -297,63 +297,6 @@ def test_unfinished_file():
                         "Hit end of file while consuming a neurite",
                         ":12:error")
 
-
```
wizmer, line 300: This portion has been moved in the test file of the mutable API.

src/morphology.cpp
```diff
@@ -34,7 +34,6 @@ Morphology::Morphology(const Property::Properties& properties, unsigned int opti
     // their respective loaders
     if (properties._cellLevel.fileFormat() == "h5") {
         mut::Morphology mutable_morph(*this);
-        mutable_morph.sanitize();
```
tomdele, line 37: hopefully you did it this way via a sanitize function...

src/mut/morphology.cpp
```diff
@@ -357,12 +354,14 @@ void Morphology::write(const std::string& filename) {
...
     if (extension == ".h5")
-        writer::h5(clean, filename);
+        writer::h5(*this, filename);
     else if (extension == ".asc")
-        writer::asc(clean, filename);
-    else if (extension == ".swc")
+        writer::asc(*this, filename);
+    else if (extension == ".swc") {
+        morphio::mut::Morphology clean(*this);
+        clean.sanitize();
```
tomdele, line 362: Can you add a comment here on why we must sanitize only for swc ?

tests/test_2_neurolucida.py
```diff
@@ -389,7 +332,11 @@ def test_section_single_point():
 
 
 def test_single_children():
-    '''Single children are merged with their parent'''
+    '''Single children are no longer merged with their parent
+
+    They used to be until the decision in:
+    https://github.com/BlueBrain/MorphIO/issues/235
```
tomdele, line 338: we should add the number version if possible also.

tests/test_2_neurolucida.py
```diff
@@ -409,13 +356,7 @@ def test_single_children():
                       )
                  ''') as tmp_file:
 
-        with captured_output() as (_, err):
```
tomdele, line 412: I cannot comment on it but shouldn't you remove the ; merged with parent section from the tmp_asc_file ?

wizmer: Ah yes good point.

src/morphology.cpp
wizmer: Yep. It's already exposed but with the latest discussion I don't know if it should remain so or if we should make it private again ?

src/mut/morphology.cpp
mgeplf: I think it would make sense to, throw an exception when a format can't represent the data, and in the exception say how to fix it - that way ppl know they have a problem, and that they need to solve it.

tests/test_2_neurolucida.py
mgeplf: cool, was going to ask exactly that :)

src/morphology.cpp
```diff
@@ -50,7 +49,6 @@ Morphology::Morphology(const std::string& source, unsigned int options)
     : Morphology(loadURI(source, options), options) {}
 
 Morphology::Morphology(mut::Morphology morphology) {
-    morphology.sanitize();
```
mgeplf, line 53: would it make sense to rename `sanitize` to 'remove_unifurcations` or something similar? or does it do more?

mgeplf (commented):
I'm +2; just need to make a decision about raising an exception vs. silently fixing for SWC.

Thanks

src/morphology.cpp
tomdele: I don t think so. This will probably be used to "sanitize" the morph for the "safe zone".
In my mind we will have people doing : 
```
morph.sanitize().write()
```

src/morphology.cpp
wizmer: Done

wizmer:
CI is crashing but I can't reproduce locally. I might use the CI to debug that.

mgeplf (commented):
Let's see what the others think - if there is not response by Wednesday, we'll merge it.

NadirRoGue (approved):

include/morphio/mut/morphology.h
```diff
@@ -210,8 +210,13 @@ class Morphology
        Fixes the morphology single child sections and issues warnings
        if the section starts and ends are inconsistent
      **/
-    void sanitize();
-    void sanitize(const morphio::readers::DebugInfo& debugInfo);
+    void removeUnifurcations();
+    void removeUnifurcations(const morphio::readers::DebugInfo& debugInfo);
+
```
alexsavulescu, line 215: Maybe this would be helpful for some: 
```suggestion
bool hasUnifurcations()
```

src/mut/morphology.cpp
```diff
@@ -255,6 +255,23 @@ void Morphology::sanitize(const morphio::readers::DebugInfo& debugInfo) {
...
+
+        auto parent = section_->parent();
+        bool isUnifurcation = parent->children().size() == 1;
+
+        if (isUnifurcation) {
+            throw WriterError(readers::ErrorMessages().ERROR_ONLY_CHILD_SWC_WRITER(parentId));
+        }
+    }
+}
+
```
alexsavulescu, line 274: So this could be:
```suggestion
bool Morphology::hasUnifurcations() {
    for (auto it = depth_begin(); it != depth_end(); ++it) {
        std::shared_ptr<Section> section_ = *it;
        if (section_->isRoot())
            continue;

        unsigned int parentId = section_->parent()->id();

        auto parent = section_->parent();
        if (parent->children().size() == 1) return true;    
    }
    return false;
}
void Morphology::_raiseIfUnifurcations() {
    if (hasUnifurcations()) {
        throw WriterError(readers::ErrorMessages().ERROR_ONLY_CHILD_SWC_WRITER(parentId));
    }
}
```

src/mut/morphology.cpp
```diff
@@ -357,12 +371,13 @@ void Morphology::write(const std::string& filename) {
         extension += my_tolower(c);
 
     if (extension == ".h5")
-        writer::h5(clean, filename);
+        writer::h5(*this, filename);
```
alexsavulescu, line 374: Eventually have:
```suggestion
        _informIfUnifurcations();
        writer::h5(*this, filename);
```
to make people aware of unifurcations - spread the word :)

alexsavulescu (approved):
LGTM. Just a usability suggestion, feel free to disregard.

arnaudon:
+1 for `_informIfUnifurcations()`, but not sure on the implications with logs as we discussed.

README.md
```diff
@@ -43,8 +43,10 @@ It supports the following formats:
 It provides 3 C++ classes that are the starting point of every morphology analysis:
 - `Soma`: contains the information related to the soma.
 
-- `Section`: a section is the succession of points between two bifurcations. To the bare minimum
-the `Section` object will contain the section type, the position and diameter of each point.
+- `Section`:
+  - A succession of points between two bifurcations if the file is in SWC format.
+  - A succession of points (not necessarily between two bifurcations) in an S-exp if the file is in ASC format.
```
arnaudon, line 48: what is S-exp?

arnaudon (approved):

alexsavulescu:
> +1 for `_informIfUnifurcations()`, but not sure on the implications with logs as we discussed.

Not sure either since I don't have all usecases in mind, however writing should be done offline/1-level-up

README.md
mgeplf: A Lisp thing; the neurolucida format reminded me of it initially:

https://en.wikipedia.org/wiki/S-expression

arnaudon: I see, very geeky!

src/mut/morphology.cpp
tomdele: +1 for this indeed

wizmer:
Going in the direction "MorphIO is a low level" library, I'm more tempted to actually strip out more functionalities.

- I am not in favor of `hasUnifurcation", I think this is already belongs the analysis realm, and thus should be in [NeuroM checks](https://github.com/BlueBrain/NeuroM/blob/master/neurom/check/structural_checks.py)
- Getting rid of the public "sanitize". This is the realm of repair and should be part of [NeuroR sanitize module](https://github.com/BlueBrain/NeuroR/blob/master/neuror/sanitize.py)

Otherwise, we start duplicating functionalities in multiple tools and won't know where to look for things.

Regarding of the unifurcation logs, I don't know what to think. Some people like them and some don't.
Btw @matz-e you can disable all logs with `morphio::set_maximum_warnings(0)`.

Please let me know what you think.

matz-e:
Having all this in NeuroM sounds very reasonable. Wouldn't this also save some code duplication?

I know about that log function. I think some output would be useful, but I'm not convinced that a library dumping it on stdout is the right way. I'd prefer the "parent" application to do that.

mgeplf:
> Going in the direction "MorphIO is a low level" library, I'm more tempted to actually strip out more functionalities.

On the writing side, it's still useful to be able to remove unifurcations, I think, no?

NeuroM doesn't handle writing, and NeuroR is python only, so if we need to write morphologies with `C++`, that functionality is needed.

alexsavulescu:
For me `remove` goes hand in hand with `has` so as to encourage explicit use. I'd rather keep these functionalities in MorphIO, they are related to writing anyway.

include/morphio/mut/morphology.h
mgeplf: let's make another issue for that

src/mut/morphology.cpp
mgeplf: let's also add an issue for that.

src/mut/morphology.cpp
mgeplf: agreed; can you add that to the review?

matz-e (commented):
If @alexsavulescu is happy, I'm happy.

arnaudon:
> If @alexsavulescu is happy, I'm happy.

Looks like an happy path!

alexsavulescu:
I'm always happy :)

wizmer:
If everyone is happy, should I merge @mgeplf ?

mgeplf:
This is still pending: https://github.com/BlueBrain/MorphIO/pull/238#discussion_r594102051

wizmer:
I replied in https://github.com/BlueBrain/MorphIO/pull/238#issuecomment-793912765.
Imo, we should not add functions to the public API in prevision of a future use-case. We can always add this later if it is really needed, no (I'm not sure it will).
But up to you.

alexsavulescu:
Was just a suggestion and it's fine with me either way, in HPC we are only looking at reading :)

tomdele:
I am fine with both. This is not a big deal to add it afterwards.

mgeplf:
Ok, let's add it after, and merge this.

arnaudon:
shall we merge that? I'll update morph-tools with the complete sanitize once this this released.

mgeplf:
Yup, let's do this.

arnaudon:
Thanks! Could you make a release, so I can update morph-tool? I don't seem to have the rights to do that. Thanks!

tomdele:
@mgeplf do you think we are ready to release this ?

arnaudon:
It is just that for some reason, pip instal morphio sends me a lot of red errors. Benoit had a quick look I think, but didn't fix it apparently.

mgeplf:
> It is just that for some reason, pip instal morphio sends me a lot of red errors. Benoit had a quick look I think, but didn't fix it apparently.

Can you open a ticket w/ the OS, setup, command you're running and the backtrace?

mgeplf:
> @tomdele  do you think we are ready to release this ?

Yes

PR #241, opened 2021-03-04
WIP: Change mut::Morphology class to template mut::TMorpholgy in orde…

jacquemi-bbp:
WIP:  Change morphio::mut::Morphology class to template morphio::mut::TMorpholgy in order to prepare morphio::mut::GLialCell and morphio::mut::Spine class implementation.

This is a first step, morphio::Morphology will have to be change to template also.
After some discussion with @wizmer it seems to be the most proper solution. 

The code does not compile for the time being.


eleftherioszisis:
A small detail: Wouldn't it be a little clearer if the template parameter `Type` was replaced by `SectionType`?
```
template<typename Type>
class TMorphology
```

jacquemi-bbp:
> A small detail: Wouldn't it be a little clearer if the template parameter `Type` was replaced by `SectionType`?
> 
> ```
> template<typename Type>
> class TMorphology
> ```

SectionType is already defined as a type, but SectionT (like in iterator.hpp for example) is much clearer, I agree

jacquemi-bbp:
> A small detail: Wouldn't it be a little clearer if the template parameter `Type` was replaced by `SectionType`?
> 
> ```
> template<typename Type>
> class TMorphology
> ```

Yes, I agree. But SectionType is already an enum used to define Neuron section types.
I propose to change the enum to NeuronSectionType (and later we will define GlialSectionType ans SpineSectionType) and to used SectionType as typename for the enum.

include/morphio/mut/morphology.h
```diff
@@ -18,42 +17,56 @@
...
 namespace mut {
+
+template <typename Type>
+class TMorphology;  // pre-declare the template class itself
+template <typename Type>
+bool diff(const TMorphology<Type>& left,
+          const TMorphology<Type>& right,
+          morphio::enums::LogLevel verbose);
+
+// template<typename Type> void modifiers::nrn_order(TMorphology<Type>& morpho);
```
jacquemi-bbp, line 31: I forget to remove this line.

PR #242, opened 2021-03-08
WIP template mut::TMorpholgy

jacquemi-bbp:
- Change morphio::mut::Morphology to template in order to prepare GlialCell and Spine classes


jacquemi-bbp:
Will be replaced by another PR

Issue #243, opened 2021-03-09
Fix Compilation error/warning morphio/mut/writers.cpp(378): error #3291: invalid narrowing conversion from "morphio::enums::CellFamily" to "unsigned int"


mgeplf:
Depending on the compiler, this is firing as an error (intel/msvc), so it corrected.

mgeplf:
That would be good.

An alternate way around that would be to have the `MorphoKit` jobs that we discussed do it.

PR #246, opened 2021-03-09
Make intel compiler happy

wizmer:
By fixing a narrow conversion
Fixes: #243

### DIFF
--- a/src/mut/writers.cpp
+++ b/src/mut/writers.cpp
@@ -375,7 +375,9 @@ void h5(const Morphology& morpho, const std::string& filename) {
     HighFive::Group g_metadata = h5_file.createGroup("metadata");
 
     write_attribute(g_metadata, "version", std::vector<uint32_t>{1, 2});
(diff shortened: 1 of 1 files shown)


mgeplf (approved):

GianlucaFicarelli:
> I guess this works for now, but the enum's type should probably be made explicit at some point in the future. I'll make a ticket, but we don't need to change it now.

As an alternative to casting, it seems that the error could be fixed adding `unsigned int` to the enum definition [here](https://github.com/BlueBrain/MorphIO/blob/78de8256f18ff7d7f384ad0b028b22f00373b54d/include/morphio/enums.h#L43), as in:
```cpp
enum CellFamily : unsigned int { NEURON = 0, GLIA = 1 };
```
Is this what you suggest?

mgeplf:
> Is this what you suggest?

Yeah, that's what I mentioned in https://github.com/BlueBrain/MorphIO/issues/247#issue-825996087

It's a breaking ABI change though, so if we're going to do it properly, we need to do it all at once, and version the ABI.

PR #248, opened 2021-03-10
[NSETM-1391] placeholder for readthedocs

asanin-epfl:


mgeplf:
Can you have the directory layout be similar to libsonata:
```
|-- environment.yml                                      
|-- Makefile                                             
`-- source                                               
    |-- *.rst                                                                                                                         
    |-- conf.py 
```

doc/source/index.rst
```diff
@@ -0,0 +1,9 @@
+.. include:: ../../README.rst
+   :end-before: .. substitutions
+
+.. toctree::
+   :hidden:
+
+   Home <self>
+   specification
```
asanin-epfl, line 8: as you can see `api` and `cpp` pages are not built

doc/source/DoxygenLayout.xml
```diff
@@ -0,0 +1,226 @@
+<doxygenlayout version="1.0">
```
mgeplf, line 1: does this work w/o doxygen? or is it needed for pybind11 mkdocs?

asanin-epfl: it does work without doxygen but only because I disabled autodoc generation as I mentioned here https://github.com/BlueBrain/MorphIO/pull/248#discussion_r592468748. This PR is a placeholder only. It does not generated proper documentation from sources. Should I also include here autodoc generation as well?

mgeplf: > This PR is a placeholder only

Yup, that's fine.

I think we should remove the doxygen stuff for now, and add it to the PR when proper documentation is generated.

asanin-epfl: I made it but CI is failing due to unrelated reasons. I will try to rerun CI tomorrow.

mgeplf (approved):

asanin-epfl:
Thank you!

Issue #218, opened 2020-12-17
A split at the root point should produce two distinct neurites.


wizmer:
The following code should pass:

```python
def test_root_node_split():
    '''Test that a bifurcation at the root point produces two root sections.'''
    with tmp_asc_file('''
                     ((Dendrite)
                      (3 -4 0 2)
                      (
                        (3 -10 2 4)
                       |
                        (3 -10 0 2)
                      )
                     )''') as tmp_file:

        n = Morphology(tmp_file.name)
        nt.assert_equal(len(n.root_sections), 2)
```

This issue appeared in @elisabettai 's case: /gpfs/bbp.cscs.ch/project/proj55/data/morphologies/mouse/asc/AA0089.asc

The problem of keeping this neurite as a single entity is that the first section end up being made of only one point, which is forbidden:
Currently the first section is giving:
```
n.section(0).points: [[ 3. -4.  0.]]
```
CC: @arnaudon

wizmer:
I should also add that this issue is also related to our decision to not automatically connect the first point of a neurite to the soma.
If this would have been the case, the first section would be composed of the soma interpolated point followed by [3, -4, 0].
As a reminder, the rational for this decision is that interpolating the crossing of the neurite and the soma is somewhat arbitrary so it would not suit everyone either.

elisabettai:
Just a minor addition to that specific morphology.

I generated it (a while ago) by using morph-tool conversion utility. The original file is /gpfs/bbp.cscs.ch/project/proj55/data/morphologies/mouse/swc/AA0089.swc

wizmer:
@tomdele I've assigned you this one because this is an easy one (I think) and it deals with the internals of the code, so I think it's a good starting point for the take over. We can pair program if you want.

tomdele:
Due to https://github.com/BlueBrain/MorphIO/issues/218, this will not be done for the moment. The PR is still here 
https://github.com/BlueBrain/MorphIO/pull/249
if we need it at some point.

Thanks to this we fixed problems in https://github.com/BlueBrain/MorphIO/pull/251.

PR #249, opened 2021-03-11
Fix split at the root point #218

tomdele:


tomdele:
This will not be used due to #238

PR #251, opened 2021-03-12
Fix deleteSection and appendSection

tomdele:
* Remove the shared pointer const refs from the methods which change the shared pointer ref count in vectors.
* Fix the deleteSection method. Problems with removal of the rootSection with recursive at false due to the behavior of the c++ map [] accessor when a key is missing (for the rootSection).


src/mut/morphology.cpp
```diff
@@ -176,14 +175,16 @@ void Morphology::deleteSection(const std::shared_ptr<Section>& section_, bool re
             deleteSection(*it, false);
         }
     } else {
-        for (auto& child : section_->children()) {
-            // Re-link children to their "grand-parent"
-            _parent[child->id()] = _parent[id];
-            _children[_parent[id]].push_back(child);
-            if (section_->isRoot())
+        // careful not to use a ref here or you will face ref invalidation problem with vector resize
```
wizmer, line 178:
```suggestion
        // Careful not to use a reference here or you will face ref invalidation problem with vector resize
        // The argument `section_` of this function could be a reference to the `_rootSections` array 
        // and that reference might be invalidated by this `push_back` (in case `vector` needs to reallocate the array)
```

src/mut/section.cpp
```diff
@@ -89,7 +89,7 @@ std::ostream& operator<<(std::ostream& os, const std::shared_ptr<Section>& secti
     return os;
 }
 
-std::shared_ptr<Section> Section::appendSection(const std::shared_ptr<Section>& original_section,
+std::shared_ptr<Section> Section::appendSection(std::shared_ptr<Section> original_section,
```
wizmer, line 92: Can you add this comment at line 113 (GH does not allow comments on unchanged lines...):

```
        // Careful not to use a reference here or you will face ref invalidation problem with vector resize
        // The argument `original_section` of this function could be a reference to the `_children` array 
        // and that reference might be invalidated by this `push_back` (in case `vector` needs to reallocate the array)
```

wizmer (approved):

PR #256, opened 2021-03-17
Template morphology glialcell

jacquemi-bbp:


include/morphio/ttree.tpp
```diff
@@ -0,0 +1,318 @@
...
+        res.push_back(static_cast<typename Node::Type>(type));
+    return res;
+}
+
+/*
+template <typename Node, typename CRTP, typename Mut>
+const CellFamily& TTree<Node, CRTP, Mut>::cellFamily() const {
+    return _properties->cellFamily();
+}
+*/
```
wizmer, line 284: dead code

src/mut/glial_cell.cpp
```diff
@@ -1,19 +1,350 @@
+#include <assert.h>
```
wizmer, line 1: For other reviewers: the factorisation of the mutable API using the TTRee concept will be done in another review.

include/morphio/types.h
```diff
@@ -52,4 +60,22 @@ using MorphologyVersion = std::tuple<std::string, uint32_t, uint32_t>;
 template <typename T>
 using range = gsl::span<T>;
 
+
+struct CellFamily {
+    struct NEURON {
+        using Type = SectionType;
+        static constexpr int value = 0;
```
wizmer, line 67: Those values must correspond to the values of the cell family in the spec.

Maybe you can add a comment about it in the code.

include/morphio/types.h
```diff
@@ -52,4 +60,22 @@ using MorphologyVersion = std::tuple<std::string, uint32_t, uint32_t>;
...
+        using Type = GlialSectionType;
+        static constexpr int value = 1;
+    };
+};
+
+using NeuronalSection = Node<CellFamily::NEURON>;
+using GlialSection = Node<CellFamily::GLIA>;
+
+// legacy name
+using Section = NeuronalSection;
```
wizmer, line 79: Note for the future. I think it would be best to stop using `Section` and to start using `NeuronalSection` when talking about neuronal sections.
I don't know how could we have a deprecation of `Section` in C++. Something like we would do in python:

```
"Section" is deprecated, please use "NeuronalSection" when refering to neuronal section
```

setup.py
```diff
@@ -75,7 +75,8 @@ def build_extension(self, ext, cmake):
         if self.cmake_defs:
             cmake_args += ["-D" + opt for opt in self.cmake_defs.split(",")]
 
-        cfg = 'Debug' if self.debug else 'Release'
+        # cfg = 'Debug' if self.debug else 'Release'
+        cfg = 'Debug'
```
wizmer, line 79: To be removed.

binds/python/bind_immutable.cpp
```diff
@@ -149,14 +149,130 @@ void bind_immutable_module(py::module& m) {
             "- morphio.IterType.breadth_first (default)\n"
             "iter_type"_a = IterType::DEPTH_FIRST);
 
-    py::class_<morphio::GlialCell, morphio::Morphology>(m, "GlialCell")
-        .def(py::init<const std::string&>())
-        .def(py::init([](py::object arg) {
-                 return std::unique_ptr<morphio::GlialCell>(new morphio::GlialCell(py::str(arg)));
+    py::class_<morphio::GlialCell>(m, "GlialCell")
```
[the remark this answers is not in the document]
jacquemi-bbp: After having discussed this point with Benoit, it appears that it is not doable because different Type are also used morphio::Morphology vs morphio::GlialCell for example.

PR #257, opened 2021-03-18
fix rstrip of documentation

asanin-epfl:


mgeplf (approved):
nice - does libsonata need this too?

asanin-epfl:
libsonata is lucky enough to miss those characters at the end but yes. I will do a similar PR there as well.

PR #258, opened 2021-03-18
Fix the mutable python tests

tomdele:
The python test: `test_lifetime_destroyed_morphology` appeared twice in the tests and so the first one was overrode and not run.


asanin-epfl (approved):

tomdele:
Thank you !

mgeplf:
nice catch, thanks

PR #263, opened 2021-03-23
Adding C++ tests

tomdele:


tests/test_mutable_morphology.cpp
```diff
@@ -0,0 +1,59 @@
+#include "contrib/catch.hpp"
+
+#include <morphio/morphology.h>
+#include <morphio/mut/morphology.h>
+
+#if defined(__APPLE__)
```
mgeplf, line 6: what if the *tests* (and only the tests) c++17, so that `experimental` isn't used, and std::filesystem is available?

tomdele: I would be VERY happy :D 

But does it makes sense to test the code in 17 when we release it in 11 ?

tomdele: I thought about it but it was not the best solution imao so that s why I tried to fix this with 11. (It was fine with linux but my mac is broken so this is a nightmare for the mac compilation)

tomdele:
Reminder for me : 
- Can use c++17 and <filesystem> if cmake > 7 --> tests ok but lcov fails
- Cannot use lcov with cmake > 7
- lcov cannot be apt-installed for version > 1.13
- prb need lcov > 1.15 with cmake > 7
- Should add -DLIBCXX_INSTALL_FILESYSTEM_LIBRARY=YES for macOS

mgeplf:
> Can use c++17 and if cmake > 7 --> tests ok but lcov fails

Can you point me to that in log? cmake shouldn't interfere w/ lcov, so it's a config issue on my side, prob.

tomdele:
> Can you point me to that in log? cmake shouldn't interfere w/ lcov, so it's a config issue on my side, prob.

@mgeplf you can have a look here : 
https://github.com/BlueBrain/MorphIO/pull/263/checks?check_run_id=2218887543

Probably related to : https://github.com/linux-test-project/lcov/issues/38

tomdele:
Ok so for the coverage I managed to run it with c++17 with `lcov1.15` and `gcov` compatible with gcc8/9. 

So I am using : 
* gcc9 
* lcov from :  wget "https://github.com/linux-test-project/lcov/releases/download/v1.15/lcov-1.15.tar.gz"
* gcov9

tests/CMakeLists.txt
```diff
@@ -1,15 +1,23 @@
...
+else()
+  set(TESTS_LINK_LIBRAIRIES ${TESTS_LINK_LIBRAIRIES} PUBLIC stdc++fs)
+endif()
+
 add_executable(unittests ${TESTS_SRC})
 
 set_target_properties(unittests
   PROPERTIES
-  CXX_STANDARD 11
+  CXX_STANDARD 17
```
mgeplf, line 20: please add a comment saying that 17 is only for tests, and it's for access to std::filesystem, in particular

.github/workflows/coverage_test.yaml
```diff
@@ -16,7 +19,21 @@ jobs:
             - name: Get submodules
               run: git submodule update --init --force --recursive
             - name: Install packages
-              run: sudo apt-get update && sudo apt-get install build-essential libhdf5-dev lcov
+              run: sudo apt-get update && sudo apt-get install build-essential libhdf5-dev
+            - name: Install lcov
+              run: |
+                VERSION="1.15"
+                wget "https://github.com/linux-test-project/lcov/releases/download/v${VERSION}/lcov-${VERSION}.tar.gz"
```
mgeplf, line 26: huh, I wonder why this worked on my `g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0` system

.github/workflows/coverage_test.yaml
```diff
@@ -16,7 +19,21 @@ jobs:
             - name: Get submodules
               run: git submodule update --init --force --recursive
             - name: Install packages
-              run: sudo apt-get update && sudo apt-get install build-essential libhdf5-dev lcov
+              run: sudo apt-get update && sudo apt-get install build-essential libhdf5-dev
+            - name: Install lcov
+              run: |
+                VERSION="1.15"
```
mgeplf, line 25: can you add a comment why we can't use the default one?

tests/test_mutable_morphology.cpp
```diff
@@ -0,0 +1,56 @@
+#include "contrib/catch.hpp"
+
+#include <morphio/morphology.h>
+#include <morphio/mut/morphology.h>
+
+#include <filesystem>
+// keep it as a namespace for simplicity because the filesystem is not really stable for all
```
mgeplf, line 7: don't need this comment, imo

tests/test_mutable_morphology.cpp
```diff
@@ -0,0 +1,56 @@
...
+                                                                               {0, {1, 2}},
+                                                                               {3, {4, 5}}};
+
+    REQUIRE(morph.connectivity() == expectedConnectivity);
+}
+
+
+TEST_CASE("writing", "[mutableMorphology]") {
+    morphio::mut::Morphology morph("data/simple.asc");
+    fs::create_directories("tmp_files");
```
mgeplf, line 45: prob. should create the files in `std::filesystem::temp_directory_path`, since there's no guarantee they'll be cleaned up

tests/test_immutable_morphology.cpp
```diff
@@ -0,0 +1,386 @@
+#include "contrib/catch.hpp"
+#include <cmath>
+#include <gsl/gsl>
```
mgeplf, line 3: this needed?

tests/test_immutable_morphology.cpp
```diff
@@ -0,0 +1,386 @@
...
+                               0.01));
+    REQUIRE(rootSection.children().empty());
+}
+
+
+TEST_CASE("endoplasmic_reticulum", "[immutableMorphology]") {
+    morphio::Morphology morph = morphio::Morphology("data/h5/v1/endoplasmic-reticulum.h5");
+    morphio::EndoplasmicReticulum er = morph.endoplasmicReticulum();
+    REQUIRE(er.sectionIndices() == std::vector<uint32_t>{1, 4, 5});
+#ifdef MORPHIO_USE_DOUBLE
```
mgeplf, line 304: can you avoid the `#ifdef` by constructing the expected values with `morphio::floatType`?

tests/CMakeLists.txt
tomdele: done

.github/workflows/coverage_test.yaml
tomdele: Really ? Because it failed on both my PC and actions with gcc>7 (ubuntu18 tho). Do you have a lcov1.15 on ubuntu20 ?

.github/workflows/coverage_test.yaml
tomdele: sure

tests/test_mutable_morphology.cpp
tomdele: yes sure this is an artifact from previous tests I made and forgot to remove

tests/test_mutable_morphology.cpp
tomdele: Sure

tests/test_immutable_morphology.cpp
tomdele: nope removed.

tests/test_immutable_morphology.cpp
tomdele: No due to the Werror + Wdouble-promotion + Wuseless-cast altogether I am quite force to do that. 

With gcc, if I am using  MORPHIO_USE_DOUBLE then morphio::floatType(0.42f)  works but morphio::floatType(0.42) throws a Wuseless-cast.

If I am not using  MORPHIO_USE_DOUBLE then morphio::floatType(0.42f) throws a Wuseless-cast  the but works with morphio::floatType(0.42). 

And I did not want to remove the warnings.

mgeplf: ok, I guess we can templetize it then, sigh

.github/workflows/coverage_test.yaml
mgeplf: I have:
```
lcov: LCOV version 1.14
```
IIRC, the github actions runner's `latest` tag for ubuntu is 20.04, so maybe we should move to that, if it removes these headaches.

It can be done in another PR.

tests/test_immutable_morphology.cpp
tomdele: I am not sure what you want to templetize here

.github/workflows/coverage_test.yaml
tomdele: Yeah I think we should keep it like this atm since it is done and check afterwards

tomdele:
OK I think we should merge as it is so at least we have a working version. Then if needed we will update.

tomdele:
This looks nice and clean now. Thanks @mgeplf for the help. 
Guess we can merge. @mgeplf, do you want to add something else ?

.github/workflows/coverage_test.yaml
```diff
@@ -1,10 +1,13 @@
...
 on: [pull_request, push]
+env:
+  CC: gcc-9
+  CXX: g++-9
 
 jobs:
     build:
         name: coverage-test
-        runs-on: ubuntu-18.04
+        runs-on: ubuntu-20.04
```
tomdele, line 10: what s the lcov version with 20 ? 1.14.2 ?

tests/test_immutable_morphology.cpp
```diff
@@ -0,0 +1,356 @@
...
+#include <morphio/mito_section.h>
+#include <morphio/mitochondria.h>
+#include <morphio/morphology.h>
+#include <morphio/mut/morphology.h>
+#include <morphio/properties.h>
+#include <morphio/section.h>
+#include <morphio/soma.h>
+
+
+namespace {
```
tomdele, line 15: just for my knowledge : in this particular case using static or an anonymous namespace is identical no ?

mgeplf: yeah; anonymous namespace is more `C++` ish; I still do `static` sometimes when I'm destracted.

mgeplf (approved):

.github/workflows/coverage_test.yaml
mgeplf: at least on my install, yes

tomdele:
Thanks mike

Issue #262, opened 2021-03-23
Introduce a flag/constructor argument to raise warning as errors instead of printing them


asanin-epfl:
Currently there is no proper way to detect warnings programmatically from MorphIO except parsing stderr. Would be good to raise errors or somehow wrap those warning as Python warnings. I have no preference. Anything to catch them programmatically on Python side.

tomdele:
I don't think we can wrap these as python warnings. The warnings are just printed strings and we do not throw / return anything.
We can maybe add a flag to throw instead but this is a bit weird. I need to think about it

asanin-epfl:
> We can maybe add a flag to throw instead but this is a bit weird. 

This will work for me

mgeplf:
> The warnings are just printed strings

A library writing output seems weird to me; it should be programatically handle-able.

There are only a few files that are writing to `cout`/`cerr`:
```
src/properties.cpp
include/morphio/section_base.tpp
src/errorMessages.cpp
src/vasc/section.cpp
```
Those should be cleaned up.

Note that `morphio.set_maximum_warnings(int)` allows one to set the errors printed; and iirc, there is a way to get a log of the errors, but I don't see it in a quick pass of the doc.

asanin-epfl:
> Those should be cleaned up.

How to clean up? I see a min and a max solution. The min is to rename `void printError`  to `void raiseError`, raise instead of printing inside this function. The max solution is to create an appropriate Exception class hierarchy because currently exceptions are functions(one function per exception type) with a tangled mix of `enum Warning` and `enum ErrorLevel`.

PR #266, opened 2021-03-31
Minimal solution to raising of warning instead of writing them

asanin-epfl:
For presentation purpose only. For #262

### DIFF
--- a/include/morphio/errorMessages.h
+++ b/include/morphio/errorMessages.h
@@ -13,7 +13,7 @@ namespace morphio {
    Set the maximum number of warnings to be printed on screen
 **/
 void set_maximum_warnings(int n_warnings);
-void raise_warnings(bool is_raise);
+void set_raise_warnings(bool is_raise);
 void set_ignored_warning(Warning warning, bool ignore = true);
 void set_ignored_warning(const std::vector<Warning>& warning, bool ignore = true);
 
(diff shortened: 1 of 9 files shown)


tomdele:
If we choose this path (that is, allowing to raise if warnings) I am fine with the implementation. 
Just the names should change.

asanin-epfl:
@mgeplf , @tomdele this one is ready for review. I can't edit coverage plan. It fails due to 0.01% decrease.

include/morphio/errorMessages.h
```diff
@@ -13,6 +13,7 @@ namespace morphio {
    Set the maximum number of warnings to be printed on screen
 **/
 void set_maximum_warnings(int n_warnings);
+void raise_warnings(bool is_raise);
```
tomdele, line 16: for homogeneity I think you should name it : 
`set_raise_warnings`

asanin-epfl: indeed, thank you!

asanin-epfl:
I will be waiting until #263 is merged, so we will have better tests here.

tomdele:
I think you can approve the #263 we should be good

asanin-epfl:
too much responsibility for me, I rather wait for Mike's final response.

tomdele (approved):

mgeplf:
Looks great, thanks for moving forward on this.

Issue #265, opened 2021-03-29
SWC files with somas w/ 2 points are of type  SomaType.SOMA_UNDEFINED; should be SOMA_CYLINDERS


mgeplf:
An SWC with:
```
1 1 0   0 0 20 -1
2 1 0 -10 0 20  1
```
 has the wrong type:
```
In [1]: import morphio
In [2]: m=morphio.Morphology('two_point_soma.swc')
In [3]: m.soma_type
Out[3]: SomaType.SOMA_UNDEFINED
In [4]: m.soma
Out[4]: <morphio._morphio.Soma at 0x7faf8c877b70>
In [5]: m.soma.points
Out[5]:
array([[  0.,   0.,   0.],
       [  0., -10.,   0.]], dtype=float32)
In [6]: m.soma.diameters
Out[6]: array([40., 40.], dtype=float32)
```

SWC always describes cylinders (w/ the weird edge case of a single point), so this should be fine: simple a cylinder w/ endopints at (0, 0, 0) <-> (0, -10, 0) and a constant diameter.

asanin-epfl:
Should we consider it as error if we agreed that MorphIO is just a reader. It does not do any processing of data it reads?

asanin-epfl:
There are other issues with MorphIO. Should we create separate tickets for them as well?
- MorphIO allows to have no soma in morphology. It only warns if there is no soma. This warning cant be caught programmatically:
  ```python
    s = load_neuron(StringIO(u"""
    1 2 0.0 0.0 0.0 0.1 -1
    2 2 0.0 0.2 0.0 0.1 1
    3 2 0.1 0.1 0.0 0.1 2
    4 2 0.0 0.0 0.0 0.853512884994 1
    9 2 3.0 6.0 0.0 0.853512884994 1
    6 2 -0.230907087225 -0.58831961861 0.257625320211 0.542229087354 5
    5 2 1.0 0.0 0.0 0.653512884994 1
    7 2 -1.07548305546 -0.79962624501 0.770244808109 0.158341330171 6
                """), reader='swc')
    print(s.soma.points)
    print(type(s.soma))
  ```
  ```
    []
    <class 'neurom.core._soma.SomaSimpleContour'>
    
    /var/folders/qy/d2s4f8916yd63kptgvzw2gxmzsmb4j/T/neurom-ei2r9qy089d410db-fd0e-41c7-8b89-f9332f971b48.swc:2:warning
    Warning: found a disconnected neurite.
[...]

PR #270, opened 2021-04-06
warn about 2-point swc soma

asanin-epfl:
Solution to #265

### DIFF
--- a/src/errorMessages.cpp
+++ b/src/errorMessages.cpp
@@ -326,7 +326,8 @@ std::string ErrorMessages::WARNING_SOMA_SWC_2POINT(const morphio::Points& points
     std::stringstream ss;
     ss << "Warning: SWC soma has 2 points. Implicitly 3-points and 1-point SWC somas\n"
           "are cylinders. 2-points SWC soma has no implicit well-established meaning.\n"
-          << points[0] << "\n" << points[1];
+       << points[0] << "\n"
+       << points[1];
     return errorMsg(0, ErrorLevel::WARNING, ss.str());
 }
 
--- a/src/errorMessages.cpp
+++ b/src/errorMessages.cpp
@@ -322,6 +322,14 @@ std::string ErrorMessages::WARNING_ONLY_CHILD(const DebugInfo& info,
     return errorMsg(0, ErrorLevel::WARNING, oss.str());
 }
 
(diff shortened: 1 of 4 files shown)


tomdele (approved):
Looks good !

asanin-epfl:
Thank you!

src/errorMessages.cpp
```diff
@@ -322,6 +322,15 @@ std::string ErrorMessages::WARNING_ONLY_CHILD(const DebugInfo& info,
     return errorMsg(0, ErrorLevel::WARNING, oss.str());
 }
 
+std::string ErrorMessages::WARNING_SOMA_SWC_2POINT(const morphio::Points& points) {
+    std::stringstream ss;
+    ss << "Warning: SWC soma has 2 points. Implicitly 3-points and 1-point SWC somas\n"
+          "are cylinders. 2-points SWC soma has no implicit well-established meaning.\n"
```
mgeplf, line 328: that's not true; SWC somas are sections, which are well defined for 2 points - I'm in the process of getting a question ready for NEURON about this.

asanin-epfl: Should I revert the PR?

tomdele: Sorry I did not know that ...

asanin-epfl: Don't be sorry. It's not well established knowledge. For example here is the old discussion about SWC with Ted https://www.neuron.yale.edu/phpBB/viewtopic.php?p=18355#p18355. The main point:

>  There has never been an "official" definition of the SWC format, so it's hard to be dogmatic about this, but "one point somas" did not start to appear until some time (years?) after SWC was already in wide use.

mgeplf: Agreed, no need to be sorry - this is a weird corner case.

IMO, it's fine to leave it until we have some clarity from what at least NEURON believes is true...which I'm trying to do.

asanin-epfl: As I pointed in https://github.com/BlueBrain/MorphIO/issues/265#issuecomment-810383942 Neuron 7.8.2 implicitly drops the 2nd point and interprets 2-point soma as a single point one. Now I've installed the latest Neuron with #907 patch. I can confirm that the same example show that Neuron does not do such implicit processing anymore. 2-point soma stays as 2-point soma. All its points stay the same.

Issue #264, opened 2021-03-24
Custom type of neurite in swc morphology


asanin-epfl:
Currently MorphIO throws an error `morphio._morphio.RawDataError` when there is a custom type but we should not decide for user what type is allowed. [SWC spec](http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html) does not restrict custom types.

@tomdele please share your thoughts. Can we replace it with a warning at least? I can make a PR here.

```
# Use custom type: 7

 1 1  0  0 0 1. -1
 2 3  0  0 0 1.  1
 3 3  0  5 0 1.  2
 4 3 -5  5 0 0.  3
 5 3  6  5 0 0.  3
 6 7  0  0 0 1.  1
 7 7  0 -4 0 1.  6
```

tomdele:
I think the problem here is that we need to have a corresponding enum <name, value> for each type. We cannot not tell the user which types of section he is using if we just put a number or a generic enum value.

tomdele:
Is this a use case at the moment ?

asanin-epfl:
It's a minor usecase for NeuroM v2. Also in general it contradicts the spec (or at least what Google gives you as the spec). Spec allows any custom neurite type. "integer value, generally continuous, starting from '1',
though this is not required." (c).

We can postpone here but it is required to fix, that's for sure.

mgeplf:
Fixed in #274

PR #274, opened 2021-04-08
Consider custom section types from 5 up to 10

asanin-epfl:
according to the spec http://neuromorpho.org/StdSwc1.21.jsp.

Solution to #264.


src/readers/morphologyHDF5.cpp
```diff
@@ -264,7 +264,7 @@ int MorphologyHDF5::_readSections() {
         const auto& p = vec[i];
         const int& type = p[1];  // Explicit int for Clang (<0 comparison)
 
-        if (type > SECTION_CUSTOM_START || type < 0) {
+        if (type >= SECTION_CUSTOM_START || type < 0) {
```
sergiorg-hpc, line 267: Good catch 👍

sergiorg-hpc:
Looks good! Before approving, I'd prefer to wait for Tom and Mike's opinion, if you don't mind.

include/morphio/enums.h
```diff
@@ -62,8 +62,8 @@ enum SectionType {
     SECTION_GLIA_ENDFOOT = 3,
 
     // All section types equal or above this number are custom types according
-    // to neuromorpho.org standard
-    SECTION_CUSTOM_START = 5,
+    // to neuromorpho.org standard (http://neuromorpho.org/StdSwc1.21.jsp)
```
tomdele, line 65: maybe there is something I did not get right. On the StdSwc1 they say: 
```
Codes 5-10 signify custom tags, typically defined in the header of the swc file when used.
In general, soma (tag 1) must start the cell and can only connect to itself. 
```
So the original  enums was ok to me no ?

asanin-epfl: Yes, the problem is that MorphIO raises an error in case of custom tag. Should MorphIO raise if custom tags are allowed tags in morphology?

tomdele: Ah ok. Hum... From what I read the correct behavior should be : 
If the custom *is not* inside the header then we raise. Otherwise we just propagate the type. 
But the problem is : in section, the type is defined using SectionType which is an enum value to simplify the queries and the bindings. But here, we are dealing with values that should be evaluated at run time and this does not leave well with the enums.

So I don t know what to do ... Maybe having enum values :
```
SECTION_CUSTOM_5 = 5,
SECTION_CUSTOM_6 = 6,
...
SECTION_CUSTOM_10 = 10,
```
could make sense ?

mgeplf: I don't think we should parse the header, I haven't seen that used, and until someone asks for it, let's skip it.

I think @tomdele suggestion of having the `SECTION_CUSTOM_5 = 5,`... is the best we can do.

asanin-epfl: well, the description is vague (as always) : "Codes 5-10 signify custom tags, **typically defined** in the header of the swc file when used." (c). This **typically defined** confuses me currently. Ans what if it is not typically defined? :D

tomdele: **typically defined (c)** == do whatever you want...

asanin-epfl: Also can you please give an example of .swc with header where custom type is defined. I've never seen how a proper header in .swc looks like.

tomdele: I think we should just create the custom tags : 
```
SECTION_CUSTOM_5 = 5,
SECTION_CUSTOM_6 = 6,
...
SECTION_CUSTOM_10 = 10,
```
without parsing the header and set the undefined to 11.

asanin-epfl: ok, thank you so much for explanation!

tomdele:
```
SECTION_OUT_OF_RANGE_START = 11
```
and raise if : 
```
type> SECTION_OUT_OF_RANGE_START 
```

asanin-epfl: Should I still add the below in case we have `SECTION_OUT_OF_RANGE_START` ?
```
    SECTION_CUSTOM_5 = 5,
    SECTION_CUSTOM_6 = 6,
    SECTION_CUSTOM_7 = 7,
    SECTION_CUSTOM_8 = 8,
    SECTION_CUSTOM_9 = 9,
    SECTION_CUSTOM_10 = 10,
```

tomdele: yes I think so. If the user want to access the custom sections, he/she will still be able to do it using the enum (which is the default way of doing it).

src/readers/morphologyHDF5.cpp
asanin-epfl: btw should we also check for `type < 0` ? `0` is not explicitly defined type. The spec does not mention it. @tomdele, @mgeplf ?

include/morphio/enums.h
asanin-epfl: thank you

src/readers/morphologyHDF5.cpp
tomdele: If not mentioned, let's consider this as invalid.

tomdele:
You should also add the section types to the bindings in bind_misc.cpp

include/morphio/enums.h
```diff
@@ -61,9 +61,15 @@ enum SectionType {
     SECTION_GLIA_PROCESS = 2,     // TODO: nasty overload there
     SECTION_GLIA_ENDFOOT = 3,
 
-    // All section types equal or above this number are custom types according
-    // to neuromorpho.org standard
-    SECTION_CUSTOM_START = 5,
+    SECTION_CUSTOM_5 = 5,
```
tomdele, line 64: You can add a comment here about the custom section

tomdele (approved):
I am good with the changes. @mgeplf can you confirm this is ok ?

include/morphio/enums.h
asanin-epfl: please verify, let me know if it is not sufficient

tomdele: Yes, that's fine.

mgeplf:
> I am good with the changes. @mgeplf can you confirm this is ok ?

Yeah, it's great - fixing the bounds (with tests) is also a nice improvement.  Thanks @asanin-epfl.

mgeplf (approved):

asanin-epfl:
Thank you!

Issue #273, opened 2021-04-08
3 point SWC soma


asanin-epfl:
Currently the only somas, that follow the below spec, are considered correct 3-point somas:
```
// The only valid neuro-morpho soma is:
// 1 1 x   y   z r -1
// 2 1 x (y-r) z r  1
// 3 1 x (y+r) z r  1
```
According to the preparing Neuron 8.0 and https://github.com/neuronsimulator/nrn/pull/1144, it's a wrong assumption. Should we stop following the above mentioned spec?

@mgeplf @tomdele

mgeplf:
No, there are two types of 3 point somas; the neuromorpho ones (http://neuromorpho.org/SomaFormat.html) with the above format (ie: `y`, `(y - r)`, `(y + r)`) and _normal_ ones, where they are just stacked cynlinders (which is truen when https://github.com/neuronsimulator/nrn/pull/1144/files#diff-61e727d6dc48930b1ab9dffbcf39e9888837c0253780242a21c7270ddc18c596L194) doesn't pass.

asanin-epfl:
ok, for me the problem is that MorphIO always throws a warning if 3-point soma does not conform the neuromorpho standard. Should this warning be thrown if 3 point soma is a *normal* one?

mgeplf:
I see, that doesn't seem right - I guess we have to fix it; question is - what is a good heuristic, such that small imprecision in y/y-r/y+r is acceptable.  Can you try and implement it?

asanin-epfl:
> Can you try and implement it?

Yes, I will implement it.

mgeplf:
I "drew" the 1/2/3 point soma cases:
![image](https://user-images.githubusercontent.com/13150119/114144026-293a9480-9915-11eb-9171-3c5e00b0216e.png)

We should redraw this nicely, and add it to the doc w/ #275, so it's clear that is happening, and say what we handle.
Couple notes:
1) we *shouldn't* do the sphere -> cylinder 'conversion' in MorphIO, that's how NEURON interprets it, iirc.
2) We need to decide if we handle the 3 point 'split/bifurcating' case

asanin-epfl:
> We should redraw this nicely

I will use matplotlib or similar tool.

> 1. we shouldn't do the sphere -> cylinder 'conversion' in MorphIO, that's how NEURON interprets it, iirc.

[...]

PR #275, opened 2021-04-08
narrow conditions when WARNING_NEUROMORPHO_SOMA_NON_CONFORM is thrown

asanin-epfl:
Solution to #273

### DIFF
--- a/doc/source/specification.rst
+++ b/doc/source/specification.rst
@@ -23,9 +23,9 @@ does not make any sense in any of the other soma formats.
 Soma single point SOMA_SINGLE_POINT
 ***********************************
 
-If a soma is made of only one point it will be treated as a "single point soma" regardless of the
-morphology file format. The single point soma is represented as a sphere whose coordinate and
-radius are given by the soma point.
+If a soma is made of only one point it will be treated as a "single point soma" for H5 and ASC formats.
+The single point soma is represented as a sphere whose coordinate and radius are given by the soma point.
+SWC interprets such soma differently.
 
 Morphology file dependent soma type:
 ************************************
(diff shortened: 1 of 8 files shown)


src/readers/morphologySWC.cpp
```diff
@@ -206,6 +206,14 @@ class SWCBuilder
         const Sample& child1 = _children[0];
         const Sample& child2 = _children[1];
```
tomdele, line 208: I think you should explain this and put a link to the issue

src/readers/morphologySWC.cpp
```diff
@@ -206,6 +206,14 @@ class SWCBuilder
...
 
+        bool isSuited = std::fabs(child1.diameter - d) < morphio::epsilon &&
+                        std::fabs(child2.diameter - d) < morphio::epsilon &&
+                        std::fabs(child1.point[0] - x) < morphio::epsilon &&
+                        std::fabs(child2.point[0] - x) < morphio::epsilon &&
+                        std::fabs(child1.point[2] - z) < morphio::epsilon &&
+                        std::fabs(child2.point[2] - z) < morphio::epsilon;
+        if (!isSuited)
```
tomdele, line 215: Same you should explain why you don t throw

src/readers/morphologySWC.cpp
asanin-epfl: ok

src/readers/morphologySWC.cpp
asanin-epfl: ok

tomdele:
Thanks for the extra explanations/comment in code. I think this is important for MorphIO since we are all in a process of learning about the morphologies and the different formats. Having this in the code makes things much easier.

tomdele (approved):
I am good with the changes

asanin-epfl:
Thank you @tomdele. We still need to figure out what bifurcation point is, and I need to make some nice images here. The PR should not be reviewed for now

tomdele:
Ok np I removed my review. Let me know when I should redo it.

doc/source/specification.rst
```diff
@@ -23,32 +23,65 @@ does not make any sense in any of the other soma formats.
...
 <http://mathworld.wolfram.com/ConicalFrustum.html>`_. Each point describes coordinates and radius
 of a disk. Each frustums is formed by 2 consecutive disks.
 
+Soma bifurcating
+~~~~~~~~~~~~~~~~
+.. image:: images/swc/split.svg
+   :scale: 100 %
+   :alt: soma two point
+
+An exceptional case of allowed bifurcating soma. TODO this needs clarifying on when such is possible.
```
asanin-epfl, line 83: @mgeplf as mentioned in #273, we need to establish when bifurcatings are allowed.

doc/source/specification.rst
```diff
@@ -23,32 +23,65 @@ does not make any sense in any of the other soma formats.
...
-`SOMA_NEUROMORPHO_THREE_POINT_CYLINDERS <https://github.com/BlueBrain/MorphIO/blob/a60b52dfe403ef289455ee2221c1b4fce6418978/src/plugin/morphologySWC.cpp#L206>`_
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-SWC files with a soma made of 3 points will have a soma of type 3-points soma. A 3-points soma is
+Soma single point
+~~~~~~~~~~~~~~~~~
+.. image:: images/swc/single.svg
+   :scale: 100 %
+   :alt: soma single point
+
+A single point sphere soma is interpreted as a cylinder of equivalent area.
```
tomdele, line 42: I think we should be more precise about the "point". This is actually not a point but a sphere/circle --> point + a diameter. 
The "point" definition differs from the single point and the two point soma.

In the single point this is a sphere. In the 2 point soma these are 2 circles.
> two points as centers of the cylinder's bases

tomdele: Am I correct ?

asanin-epfl: I will adjust. In my opinion the whole docs require some serious cleaning. I created a ticket for that #277.

asanin-epfl: Feel free to fix my definitions. I find it hard to adjust only SWC parts without affecting H5 and ASC. In my opinion we need a good general definition, that will be further concretized for each format. Currently MorphIO does not give general definition.

tomdele:
Thanks this is much clearer now

doc/source/specification.rst
tomdele: I think this should be done in a much bigger review about docs. But this takes time and this is not the priority yet. Your current definition is great atm.

tomdele (approved):

asanin-epfl:
Thank you!

PR #276, opened 2021-04-09
fix broken links and include README to readthedocs

asanin-epfl:


tomdele (approved):
looks good. I checked the url and they seem to point to the correct target.

If CI is good, I am good too

asanin-epfl:
Thank you!

PR #280, opened 2021-04-14
Allow 2 point swc soma as a cylinder one. Solution to #265.

asanin-epfl:
solution to #265


src/readers/morphologySWC.cpp
```diff
@@ -231,9 +231,6 @@ class SWCBuilder
         case 1: {
             return SOMA_SINGLE_POINT;
         }
-        case 2: {
-            return SOMA_UNDEFINED;
-        }
```
tomdele, line 236: Why not SOMA_CYLINDERS here ?

asanin-epfl: It is a default soma type. Should I explicitly set it here?

tomdele: I think so yes.

asanin-epfl: done

src/shared_utils.hpp
```diff
@@ -20,8 +20,6 @@ floatType _somaSurface(const SomaType type,
                        const ContainerDiameters& diameters,
                        const ContainerPoints& points) {
     size_t size = points.size();
-    if (size == 0)
```
asanin-epfl, line 23: This one to discuss. Because of it, SOMA_UNDEFINED becomes unreachable further in code => morphologies without soma have soma surface 0. However how correct to have 0 surface if there is no soma at all?

tomdele: I think raising is good. This is clearer and more correct imao.

tomdele (approved):
That s good !

asanin-epfl:
Thank you!

PR #282, opened 2021-04-14
use CHECK_THROW_AS from Catch2

mgeplf:
* make sure gcov is linked in


tomdele (approved):
That s nice ! Not sure why I used the ugly stuffs in the first place, I probably missed the CHECK_THROWS_AS.

PR #285, opened 2021-04-15
cleanup morphologyHDF5

mgeplf:
* make which datasets belong to which versions more clear
* don't use try/catch to detect if datasets exist
* try and clarify where member variables are read & saved
  * removed member variables _points, _pointsDims, _sections,
    since their locality can be kept local
* fix version numbering in tests/data/h5/v1/endoplasmic-reticulum.h5
                                            mitochondria.h5
* remove unnecessary `morphio::`


src/readers/morphologyHDF5.cpp
```diff
@@ -19,48 +19,65 @@
  */
 
 #include "morphologyHDF5.h"
+#include <cassert>
```
tomdele, line 22: don t think it is used

src/readers/morphologyHDF5.cpp
```diff
@@ -70,193 +87,154 @@ Property::Properties load(const HighFive::Group& group) {
...
+    if (_properties._cellLevel.minorVersion() >= 1) {
+        _readPerimeters(firstSectionOffset);
 
-    if (_pointsDims.size() != 2 || _pointsDims[1] != _pointColumns) {
-        throw morphio::RawDataError("Opening morphology '" + _uri +
-                                    "': bad number of dimensions in 'points' dataspace.");
+        if (_properties._cellLevel.minorVersion() >= 2) {
+            _readMitochondria();
+            _readEndoplasmicReticulum();
+        }
```
tomdele, line 101: Does it means if we have a version 1.2 we will always try to read the mito/endo data even if we are opening a neuron ?

src/readers/morphologyHDF5.cpp
```diff
@@ -70,193 +87,154 @@ Property::Properties load(const HighFive::Group& group) {
...
-        try {
-            _group.getGroup(_g_root).getAttribute(_a_version).read(potential_h5v2_version);
-        } catch (const HighFive::Exception&) {
+    // if there is metadata, perhaps it's h5v1, minor version 1, 2
+    if (_group.exist(_g_metadata)) {
+        const auto metadata = _group.getGroup(_g_metadata);
+        if (metadata.hasAttribute(_a_version)) {
+            const auto attr = metadata.getAttribute(_a_version);
+
+            std::array<uint32_t, 2> versions;
```
tomdele, line 130: This is not initialized and I am not familiar with HighFive but shouldn't we init to something like : 0, 0 ?

src/readers/morphologyHDF5.cpp
```diff
@@ -70,193 +87,154 @@ Property::Properties load(const HighFive::Group& group) {
...
-            // points and diameters are PODs. Fastest to resize then assign values
-            somaPoints.resize(somaPoints.size() + section_offset);
-            somaDiameters.resize(somaDiameters.size() + section_offset);
-            for (std::size_t i = 0; i < section_offset; ++i) {
-                const auto& p = hd5fData[i];
-                somaPoints[i] = {p[0], p[1], p[2]};
-                somaDiameters[i] = p[3];
-            }
+    somaPoints.resize(somaPoints.size() + section_offset);
+    somaDiameters.resize(somaDiameters.size() + section_offset);
```
tomdele, line 183: isn't somaPoints.size() always zero by definition here ?

src/readers/morphologyHDF5.cpp
```diff
@@ -70,193 +87,154 @@ Property::Properties load(const HighFive::Group& group) {
...
-            }
-        };
+    auto& points = _properties.get<Property::Point>();
+    auto& diameters = _properties.get<Property::Diameter>();
 
-    std::vector<std::array<morphio::floatType, _pointColumns>> vec(_pointsDims[0]);
-    if (vec.size() > 0) {
-        _points->read(vec.front().data());
+    if (hasNeurites) {
+        const size_t size = (points.size() + hd5fData.size() - section_offset);
```
tomdele, line 195: same for points.size() it should/must be 0 no ?

src/readers/morphologyHDF5.cpp
tomdele: my bad missed the :
```
if (!_group.exist(_g_mitochondria)) {
        return;
    }
```
and
```
if (!_group.exist(_g_endoplasmic_reticulum)) {
        return;
    }
```

tomdele:
I made the changes I proposed in :
https://github.com/BlueBrain/MorphIO/pull/288
it also fixed the no soma problem.

If you are ok with this PR above we can merge both.

asanin-epfl:
Currently I have problems with hdf morphologies that have its soma points split among several offsets in structure:
```
HDF5 "scripts/three-point-soma-two-offset.h5" {
GROUP "/" {
   DATASET "points" {
      DATATYPE  H5T_IEEE_F32LE
      DATASPACE  SIMPLE { ( 8, 4 ) / ( 8, 4 ) }
      DATA {
      (0,0): -2, 0, 0, 1,
      (1,0): 0, 0, 0, 1,
      (2,0): 0, 0, 0, 1,
      (3,0): 2, 0, 0, 1,
      (4,0): 2, 0, 0, 1,
      (5,0): 3, 3, 0, 1,
      (6,0): 2, 0, 0, 1,
      (7,0): 4, 0, 4, 1
      }
   }
   DATASET "structure" {
      DATATYPE  H5T_STD_I32LE
      DATASPACE  SIMPLE { ( 4, 3 ) / ( 4, 3 ) }
      DATA {
      (0,0): 0, 1, -1,
      (1,0): 2, 1, -1,
      (2,0): 4, 3, 0,
      (3,0): 6, 2, 0
      }
   }
}
}
```

The 2nd half of soma is skipped and soma type is undefined.
```
three-point-soma-two-offset.h5
SomaType.SOMA_UNDEFINED
[[-2.  0.  0.]
 [ 0.  0.  0.]]
section points:
	 [2. 0. 0.]
	 [3. 3. 0.]
	 [2. 0. 0.]
	 [4. 0. 4.]
```

Should I be worried about this case when soma is represented by multiple offsets?

tomdele:
Do we have specs on that ?
Having unifurcation on Neurites --> why not, but on Somas...

To me without more specs (like we can have only one soma), this is an undefined behavior an we should raise because we cannot figure out if this is double soma morphology (one single point soma and one cylinder soma) or if this is a poorly formatted 3 points soma.

asanin-epfl:
> Do we have specs on that ?

There is [something](https://developer.humanbrainproject.eu/docs/projects/morphology-documentation/0.0.2/h5v1.html#geometric-interpretation): "Soma: there is only one soma section, however, different tools interpret them differently:" (c). I propose to throw en exception for this case as we don't throw currently on the latest master branch.

tomdele:
Ok so we throw. I will update the code

tomdele:
Added a check here to ensure a single section soma and a the soma being the first section: 
https://github.com/BlueBrain/MorphIO/pull/288/commits/f7c1c6ac359199f43d764361ceb751dbf9c09ca5

tomdele:
I put you as a reviewer but this is more to know if you have something to add or comment @asanin-epfl

asanin-epfl:
From my side looks fine. I'm working on docs update currently, so it's good for me. Also a question, should be a similar PR for the ASC reader?

src/readers/morphologyHDF5.cpp
mgeplf: nice, thanks

src/readers/morphologyHDF5.cpp
mgeplf: fixed

src/readers/morphologyHDF5.cpp
```diff
@@ -213,77 +218,68 @@ int MorphologyHDF5::_readSections() {
...
-            RawDataError("Error reading morphologies " + _uri + " soma is not the first section."));
-    }
-
-    // If the morphology has a soma, it is the first section
-    bool hasSoma = !soma_sections.empty();
-    if (hasSoma && vec.size() < 2) {
+    bool hasSoma = true;
+    if (static_cast<SectionType>(vec[0][SECTION_TYPE]) != SECTION_SOMA) {
+        hasSoma = false;
+    } else if (vec.size() == 1) {
```
tomdele, line 242: indeed much better

tomdele (approved):
Looks good !

mgeplf:
I also tested it w/ NeuroM d177237, as cross check.

PR #292, opened 2021-04-22
Change glial cell section types according to new spec

eleftherioszisis:
The glial section types are changed to reflect the types that are outputted by the ngv building and synthesis. The endfoot process has also been renamed to perivascular to avoid name overlapping with the endfoot as a gliovascular connection.

Before:
```
  2 : glia process
  3 : glia endfoot
```
After:
```
  2 : glia perivascular process
  3 : glia process
```
See updated spec here:

### DIFF
--- a/binds/python/bind_misc.cpp
+++ b/binds/python/bind_misc.cpp
@@ -63,9 +63,10 @@ void bind_misc(py::module& m) {
         .value("custom8", morphio::enums::SectionType::SECTION_CUSTOM_8)
         .value("custom9", morphio::enums::SectionType::SECTION_CUSTOM_9)
         .value("custom10", morphio::enums::SectionType::SECTION_CUSTOM_10)
+        .value("glia_perivascular_process",
+               morphio::enums::SectionType::SECTION_GLIA_PERIVASCULAR_PROCESS)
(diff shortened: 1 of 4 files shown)


tomdele:
You prob need to change it also in the bindings

tomdele:
Can you add C++ and python tests just to be sure we are breaking the enum doing this and returning the correct values ?
in :
`TEST_CASE("glia", "[immutableMorphology]")` and `TEST_CASE("sections", "[immutableMorphology]")`. Or a dedicated one.

Like counting the number of SECTION_DENDRITE and SECTION_AXON and check if this is correct + checking what s happening when we use these SectionTypes on the astrocytes (from the code you should be able to use them even on a glia cell)

tomdele (approved):
ok

DexterAntonio:
Can you please provide a publically accessible link to the Glia cell specs? The link you posted is no longer publically accessible.

mgeplf:
It should be available at: https://morphology-documentation.readthedocs.io/en/latest/h5v1.html

I will update the links in the repo - thanks!

DexterAntonio:
Thank you for resolving this issue!

PR #293, opened 2021-04-22
Minor build system tweaks

sanjayankur31:
Hello,

We made a few tweaks to the build project files to adhere to the [Fedora packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/) while including [MorphIO in Fedora](https://src.fedoraproject.org/rpms/morphio). They:

- use the [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) so that files are installed in the standard locations
- install the python binding shared object to the right place
- version the shared object

We're carrying these patches in Fedora at the moment, but we'd like to [stay as close to upstream as possible](https://fedoraproject.org/wiki/Staying_close_to_upstream_projects). Could the dev team please take a look to see if these could be merged here upstream?


sanjayankur31:
> Thanks, we will have to check that this works on our side.
> 
> I don't think we'll be able to use the chosen `SOVERSION`, but you're right that it needs to be set, and we need to make a clear decision about that - appreciate it.

Thanks very much. 

I've set the SOVERSION to 0.0.0 to sort of signify that one hasn't been set (a default), so that whatever the dev team here decide will be greater than the current version.

mgeplf:
Sorry, I"m way behind and haven't had a chance to do this; it hasn't been forgotten though!

mgeplf:
I re-iterate the above comment - looking at this now.

matz-e (dismissed):
LGTM

matz-e (commented):
Building and loading `py-morphio` in Spack is fine, as is using `morphio` as a dependency for TouchDetector and running a test circuit. So from my side, good to go in.

sanjayankur31:
Hello, any progress here please? (If this gets merged, I won't have to carry and re-port these patches to new versions to update the Fedora packages, so that'll make life a little easier for us downstream).

mgeplf:
> Hello, any progress here please? (If this gets merged, I won't have to carry and re-port these patches to new versions to update the Fedora packages, so that'll make life a little easier for us downstream).

I'm so sorry, I genuinely don't want to make maintainers life harder, so we'll try and get this in for the next version we release.

What sort of timeline are you on?

mgeplf:
@sanjayankur31 thanks a lot for the change; I think we're good to go with it.

PR #296, opened 2021-04-28
Warn about disconnected neurite only when there is a soma.

asanin-epfl:
It is legal to have disconnected neurite when there is no soma. Warn only about missing soma in that case.


tomdele (approved):

asanin-epfl:
Thank you!

Issue #299, opened 2021-04-29
Throw an error when SWC has soma as contour


asanin-epfl:
If all points of SWC soma have zero radius then throw, because SWC can't have soma contour.

PR #301, opened 2021-04-30
Warn about zero diameter in SWC morphology

asanin-epfl:
@tomdele should we rather throw an error if diameter is 0? Should we check H5 and ASC as well for this?
Solution to #299


tomdele:
Hum ... I am not sure if this can be a use case or not. 
@arnaudon any insights ?

asanin-epfl:
It is a usecase when .swc is made by conversion of .asc.

tomdele:
Ok I see. 
Anyway, thinking about it, we should probably warn and not throw. I think if something is not explicitly in the specs, we should not interpret. Even if it does not make sense. As a rule of thumb, we should not create sub-specs, otherwise it will backfire at some point.

tomdele (approved):
Looks good

asanin-epfl:
Thank you!

> ... it will backfire at some point.

I agree. That's why warning is ok in this case. Also SWC is the only appropriate format because H5 and ASC can have contour somas.

tomdele:
> That's why warning is ok in this case. 

indeed it makes perfect sense here.

PR #302, opened 2021-05-03
Replace nosetests with pytest in dist publishing

asanin-epfl:


tomdele (approved):
My bad for forgetting this in the previous reviews...

asanin-epfl:
My bad I forget it as well

Issue #300, opened 2021-03-16
Support "Cell Body" for ASC files


genric:
After discussing this today, we probably want to support "Cell Body" definition format as, due to historical reasons, there are morphologies using that soma definition format and other tools are able to process them.
Example morphology:
http://neuromorpho.org/dableFiles/jonas/Source-Version/02022018_252Lgc11.ASC
available here: http://neuromorpho.org/neuron_info.jsp?neuron_name=02022018_252Lgc11)

Looks like there was a similar issue(#264) related to "Cell Body" and there was a PR to support them.

asanin-epfl:
There is an attempt to implement it already in #39.

PR #305, opened 2021-05-05
more reliable ASC (Neurolucida) parser

asanin-epfl:
- Allow neurite token to start with lower case character
- Allow soma token with whitespace: `(Cell Body)`
- Allow neurite token to be represented as text for backward compatibility with old morphologies as mentioned in #300 .


src/readers/lex.cpp
```diff
@@ -104,6 +104,13 @@ constexpr std::size_t operator+(Token type) {
     return static_cast<std::size_t>(type);
 }
 
+const std::map<std::string, Token> NeuriteStringTokenMap{{to_string(Token::AXON), Token::AXON},
```
asanin-epfl, line 107: This format is due to clang. I didn't want this mess.

src/readers/morphologyASC.cpp
```diff
@@ -40,6 +40,18 @@ bool is_neurite_type(Token id) {
             id == Token::CELLBODY);
 }
 
+std::string as_token_text(const std::string& text) {
+    std::string upp_text = text;
```
asanin-epfl, line 44: I'm open to suggestions here as I'm not C++ expert. It does:
- makes text upper case
- removes any whitespace

src/readers/morphologyASC.cpp
```diff
@@ -217,6 +225,12 @@ class NeurolucidaParser
        Parse the root sexp until finding the first sexp containing numbers
     **/
     Header parse_root_sexp_header() {
+        static const std::map<std::string, Token> NeuriteStringTokenMap{
+            {to_string(Token::AXON), Token::AXON},
```
mgeplf, line 229: Other than CellBody, do any other tokens have this same behavior?

src/readers/morphologyASC.cpp
asanin-epfl: Nope, any. it's for making `Cell Body` into `CellBody`.

mgeplf: Thanks.

src/readers/morphologyASC.cpp
mgeplf: I believe that the old style was only used for CellBody/Soma; I have reduced the scope of the PR to that.

asanin-epfl: Should I update the PR with that in mind?

src/readers/morphologyASC.cpp
```diff
@@ -217,6 +225,12 @@ class NeurolucidaParser
        Parse the root sexp until finding the first sexp containing numbers
     **/
     Header parse_root_sexp_header() {
+        static const std::map<std::string, Token> NeuriteStringTokenMap{
```
tomdele, line 228: for this usecase where we don t need stable iterators, don t care about ordering and having string as keys on a small map isn't an unordered_map a better choice ?

mgeplf: map is gone in the latest version

asanin-epfl:
Should I do anything with coverage? What is your plan here @mgeplf ?

mgeplf (approved):

PR #307, opened 2021-05-06
Cleanup properties & vasc/properties

mgeplf:
* use macro to reduce templated code
* functions exist in headers, and should be inlined in each TU
* SomaLevel was never used
* CellLevel wasn't being completely constructed


include/morphio/vasc/properties.h
```diff
@@ -68,7 +69,7 @@ struct Properties {
     std::vector<Connection::Type> _connectivity;
 
     template <typename T>
-    std::vector<typename T::Type>& get() noexcept;
+    std::vector<typename T::Type>& get_mut() noexcept;
```
tomdele, line 72: wouldn't be clearer to have : 
```
get
cget (or get_const)
```
as names instead ? I feel like the `mut` suffix collides a bit with the "mutable morphologies" naming convention. 

But this is maybe personal and a question of taste.

include/morphio/properties.h
```diff
@@ -206,59 +194,42 @@ struct Properties {
...
-template <>
-const std::vector<MitoSection::Type>& Properties::get<MitoSection>() const noexcept;
-template <>
-std::vector<MitoSection::Type>& Properties::get<MitoSection>() noexcept;
-
-template <>
-const std::vector<MitoPathLength::Type>& Properties::get<MitoPathLength>() const noexcept;
-template <>
-std::vector<MitoPathLength::Type>& Properties::get<MitoPathLength>() noexcept;
+#define INSTANTIATE_TEMPLATE_GET(T, M)                                       \
```
tomdele, line 200: are the `\` mandatory ?

tomdele (commented):
Just a question/discussion about naming otherwise it looks good.

include/morphio/properties.h
matz-e: yes

include/morphio/properties.h
```diff
@@ -206,59 +194,42 @@ struct Properties {
...
-template <>
-std::vector<MitoDiameter::Type>& Properties::get<MitoDiameter>() noexcept;
+#undef INSTANTIATE_TEMPLATE_GET
 
 template <>
-const std::vector<Section::Type>& Properties::get<Section>() const noexcept;
-template <>
-std::vector<Section::Type>& Properties::get<Section>() noexcept;
+inline const std::map<int32_t, std::vector<uint32_t>>& Properties::children<Section>() const
+    noexcept {
```
matz-e, line 224: What kind of line wrapping is this? Readability suffers a bit, I was wondering at first what a `noexcept` block is…

include/morphio/vasc/properties.h
matz-e: `const` by default seems like a better interface…

matz-e (approved):
Seems like a nice cleanup. Now if only that macro stuff could be deduplicated…

include/morphio/properties.h
mgeplf: clang-format

PR #308, opened 2021-05-10
Fix Lexertl sdist

mgeplf:
* Make sure tests pick this problem up next time.

### DIFF
--- a/.github/workflows/check-sdist.yml
+++ b/.github/workflows/check-sdist.yml
@@ -1,0 +1,27 @@
+name: Check that `sdist` builds and installs
+on: [pull_request, push]
+
+jobs:
+  build_sdist:
+    name: Build sdist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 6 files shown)


.github/workflows/check-sdist.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Build a source tarball
+        run:
+          python setup.py sdist
+
+      - name: Test tarball
+        run: |
+          sudo apt install -y libhdf5-dev
+          pip install -r tests/requirement_tests.txt
+          pip install dist/*
+          pytest -s -v tests
```
tomdele, line 27: We are doing it in the `Test tarball` from `publish-sdist-wheels.yml`. Do you think it should be done twice ?

mgeplf: they run pretty quick, so it should be fine, imo.

tomdele: I am fine with this then

tomdele (approved):

PR #313, opened 2021-05-12
Move in NeuroM specs

asanin-epfl:
Not finished. Still need to make two svg images.

### DIFF
--- a/doc/source/specification.rst
+++ b/doc/source/specification.rst
@@ -1,11 +1,57 @@
 Specification
 =============
 Description of how MorphIO sees morphologies. MorphIO supports :ref:`.asc <specification-neurolucida>` (Neurolucida),
-`.h5`_ (H5 version 1) and `.swc`_ extensions.
+`.h5`_ (H5 version 1) and `.swc`_ extensions. A morphology is represented as a soma with neurites. Either soma or
+neurites can be absent. Neurite is a tree of sections. Section is composed of segments where segment is made of points.
+Lets start describing them from the smallest to the biggest entity.
 
-Soma formats:
--------------
-MorphIO implements Soma specification from `NeuroMorpho`_, and recognizes several kinds of soma format.
+
+Point
+-----
+A point is a ``numpy`` array of three numbers **[X, Y, Z]** where X, Y, Z are Cartesian coordinates of position.
+A point always has a diameter associated with it.
+
+
+Segment
+-------
+A segment consists of two consecutive points(`Point`_) belonging to the same section(`Section`_). It is represented as
+a tuple or a ``numpy`` array of two `points<point-label>`.
+
+.. image:: images/swc/two_pt.svg
+   :scale: 100 %
+   :alt: segment
+
+
+Section
+-------
+
+A section is a series of two or more points. Each section has a type associated with it. The type shows what part of
+the neuron a section represents. The type can be the axon, the soma, and so on.
+`Section API <cpp/doxygen_Section.html>`_.
+
+.. image:: images/section.svg
+   :scale: 100 %
+   :alt: section
+
+The first and the last point of section must be of the following combinations:
+
+1. first non-soma point, forking point
+2. forking point, forking point
+3. forking point, end point
+4. first non-soma point, end point
+
+First non-soma point is the first point of a root section. Forking point is the last point of a section that has
+multiple children. End point is the last point of a section without children. For more details on the root section and
+children see `Neurite`_. The above combinations are marked with their corresponding number at the image below.
+
+.. image:: images/section_segment.svg
+   :scale: 100 %
+   :alt: section variants
+
+Soma
+----
+A soma is a series of one or more points. MorphIO implements Soma specification from `NeuroMorpho`_, and recognizes
+several kinds of soma format. `Soma API <cpp/doxygen_Soma.html>`_.
 
 No Soma
 *******
@@ -115,28 +161,17 @@ are characterized by "structure" with type equal 1.
         "structure" entries with type equal 1. Simply saying soma that is split among multiple sections is not supported
         in those formats.
 
+Neurite
+-------
 
-Sub-cellular structures
------------------------
-SWC does not support any sub-cellular structures. H5 and ASC support some, please see the following
-for more details.
+A neurite is essentially a tree of sections(`Section`_). The tree structure implies the following:
 
-Spines
-******
-SWC and H5 do not support spines. ASC files containing files will be read correctly. However
-spines data are not handled by MorphIO and spine information will be lost when writing to disk.
+* Section can have only one parent (another section).
+* Section can have an arbitrary number of children (other sections).
+* No loops are present in the structure.
 
-Mitochondria
-************
-SWC and ASC do not support mitochondria. Mitochondria can be read and written to disk using the H5
-format. See :ref:`mitochondria-readme` for more details about the mitochondria API.
-
-Custom annotations
-******************
-Custom annotations are not supported.
-
-Sections
---------
+A section without parent is called a root section. A section with parent must have its first point to be a duplicate
+of the last point of its parent.
 
 Section ordering
 ****************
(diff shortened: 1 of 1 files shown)


doc/source/specification.rst
```diff
@@ -1,11 +1,95 @@
...
+.. image:: images/swc/two_pt.svg
+   :scale: 100 %
+   :alt: segment
+
+
+Section
+-------
+
+A section is a series of one or more segments. Each section has a type associated with it. The type shows what part of
+the neuron a section represents. The type can be the axon, the soma, and so on.
```
arnaudon, line 28: could we put here the real list of sectiontypes?

doc/source/specification.rst
```diff
@@ -1,11 +1,95 @@
...
+   :alt: section
+
+The first and the last point of section must be of the following combinations:
+
+1. starting point, forking point
+2. forking point, end point
+3. forking point, forking point
+4. starting point, end point
+
+Starting point is the first point of a section without parent. Forking point is the last point of a section that has
```
arnaudon, line 42: I would move this paragraph before the bullet point, so we know what the different points represent before we read it.

doc/source/specification.rst
```diff
@@ -1,11 +1,95 @@
...
+children see `Neurite`_. The above combinations are marked with their corresponding number at the image below.
+
+.. image:: images/sections.svg
+   :scale: 100 %
+   :alt: section variants
+
+Neurite
+-------
+
+A neurite is essentially a tree of sections(`Section`_). The tree structure implies the following:
```
arnaudon, line 53:
```suggestion
A neurite is a tree of sections(`Section`_). The tree structure implies the following:
```

doc/source/specification.rst
```diff
@@ -1,11 +1,95 @@
...
+Section ordering
+****************
+In MorphIO each section is identified by an ID. By default, the section IDs will correspond to
+the order of section appearance while performing a depth-first traversal on every neurites. The
+neurite order is the order of appearance in the file. Alternatively, the NRN simulator way of
+ordering section can be used by specifying the flag ``morphio::Option::NRN_ID`` when opening
+the file. In the NRN simulator, the soma which is considered as a section (contrary to MorphIO)
+is placed first and then neurites are sorted according to their type.
+
+The final order is the following:
```
arnaudon, line 71: final in what sense? In neuron ordering?

doc/source/specification.rst
```diff
@@ -1,11 +1,95 @@
...
+
+A section without parent is called a root section. A section with parent must have its first point to be a duplicate
+of the last point of its parent.
+
+Section ordering
+****************
+In MorphIO each section is identified by an ID. By default, the section IDs will correspond to
+the order of section appearance while performing a depth-first traversal on every neurites. The
+neurite order is the order of appearance in the file. Alternatively, the NRN simulator way of
+ordering section can be used by specifying the flag ``morphio::Option::NRN_ID`` when opening
```
arnaudon, line 67: I use this: `options=morphio.Option.nrn_order`, is it the same as NRN_ID?

doc/source/specification.rst
```diff
@@ -1,11 +1,95 @@
...
+0. Soma
+1. Axon
+2. Basal
+3. Apical
+
+Section with only one child section
+***********************************
+Prior to version 3.0.0, when a section had a single child section (aka unifurcation), the child section would be merged
+with its parent when reading or writing the file. Since version 3.0.0, merging does not happen when reading. Yet
+writing of such sections is not allowed.
```
arnaudon, line 82: could you point to `remove_unifurcation` method? If you try to save, what happens, this is called? I'm not sure I tried to save with unifurcations, but I find this weird that we cannot save these as they are.

doc/source/specification.rst
asanin-epfl: Definitions go immediately after their usage. Putting them before, breaks the narration.

doc/source/specification.rst
asanin-epfl: done

doc/source/specification.rst
asanin-epfl: Done. It is funny that I took the phrase from NeuroM docs.

doc/source/specification.rst
asanin-epfl: Done

doc/source/specification.rst
asanin-epfl: done

doc/source/specification.rst
asanin-epfl: NRN_ID is a typo. Fixing this. Thank you.

arnaudon (approved):

tomdele:
I am checking now

doc/source/specification.rst
```diff
@@ -1,11 +1,96 @@
...
+   :alt: section
+
+The first and the last point of section must be of the following combinations:
+
+1. starting point, forking point
+2. forking point, end point
+3. forking point, forking point
+4. starting point, end point
+
+Starting point is the first point of a section without parent. Forking point is the last point of a section that has
```
tomdele, line 42:
```suggestion
Starting point is the first point of a section without parent inside the section. Forking point is the last point of a section that has
```

From the morphology point of view, the point has a parent but it is inside the previous section.

doc/source/specification.rst
```diff
@@ -1,11 +1,96 @@
...
+children see `Neurite`_. The above combinations are marked with their corresponding number at the image below.
+
+.. image:: images/sections.svg
+   :scale: 100 %
+   :alt: section variants
+
+Neurite
+-------
+
+A neurite is a tree of sections(`Section`_). The tree structure implies the following:
```
tomdele, line 53:
```suggestion
A neurite is an out-tree of sections(`Section`_). The tree structure implies the following:
```

Can be called an `Arborescence` also. `Tree` by definition is undirected.

doc/source/specification.rst
```diff
@@ -1,11 +1,96 @@
...
+
+* Section can have only one parent (another section).
+* Section can have an arbitrary number of children (other sections).
+* No loops are present in the structure.
```
tomdele, line 57:
```suggestion
* It exists a root section for the neurite.
* The structure is oriented : all its edges point away from the root.
* No cycles are present in the structure.
```

(In graph theory, a loop is a special case of cycles where the edge loop on the same node. Also called the self-loop)

doc/source/specification.rst
```diff
@@ -1,11 +1,96 @@
...
+
+A section without parent is called a root section. A section with parent must have its first point to be a duplicate
+of the last point of its parent.
+
+Section ordering
+****************
+In MorphIO each section is identified by an ID. By default, the section IDs will correspond to
+the order of section appearance while performing a depth-first traversal on every neurites. The
```
tomdele, line 65: I am not sure we a have a stability guaranty with DFS. Isn't the DFS a consequence of the file formats (specially for swc or ascii) ?

asanin-epfl: What is DFS? Depth-first-...? I agree that this statement is wrong. I bluntly copied it from NeuroM. I need to edit it.

tomdele: Depth-first-search sorry

tomdele: Because it depends on the ordering of the neighbors and so it depends on the graph / tree underlying implementation. If you are using something like a adjacency sets (with unordered set) you cannot guaranty a stable iterator on the tree/graph.

asanin-epfl: Can you please help me with clarification on this.  I'm lost in the sources of MorphIO.

tomdele: Yes I am checking

asanin-epfl: so, I'm putting this into docs? Would it be ok for you?

tomdele: lol actually after thinking about it and having a better understanding, I think the doc is correct...

tomdele (approved):
ok after checking I think this is quite ok ! 

Thanks for this @asanin-epfl

doc/source/specification.rst
asanin-epfl: XD, should I put something like "verified by Tom" among the lines?

asanin-epfl:
Thank you for the thorough review!

doc/source/specification.rst
tomdele: "TomDele seal of approval".

tomdele:
> Thank you for the thorough review!

That was harder than expected for me :D

PR #321, opened 2021-05-28
Dendritic spine

mgeplf:


include/morphio/enums.h
```diff
@@ -60,8 +60,16 @@ enum SectionType {
     SECTION_DENDRITE = 3,         //!< general or basal dendrite (near to soma)
     SECTION_APICAL_DENDRITE = 4,  //!< apical dendrite (far from soma)
 
-    SECTION_GLIA_PERIVASCULAR_PROCESS = 2,
-    SECTION_GLIA_PROCESS = 3,  // TODO: nasty overload there
+    SECTION_GLIA_PERIVASCULAR_PROCESS = 2,  // Note: overlaps with SECTION_AXON
+    SECTION_GLIA_PROCESS = 3,               // Note: overlaps with SECTION_DENDRITE
+
+    SECTION_SPINE_HEAD = 2,  // Note: overlaps with SECTION_AXON
```
jacquemi-bbp, line 66: The HDF5 Morphology [specification]( https://bbpteam.epfl.ch/documentation/projects/Morphology%20Documentation/latest/h5v1.html) defines the type of process for dendrite spine section as follow:
1: head, 2: neck

Is there a reason why you defined other values ?

jacquemi-bbp:
append_root_section python method is not available for morphio.mut.DendriticSpine object

include/morphio/enums.h
mgeplf: Yeah, it's a work around for now so you can test the API; the overlap w/ the soma and the different handling of it.  It will go back to the 'original' values once I have a clean workaround for it.

jacquemi-bbp:
Which method (API) is foreseen to add a post synaptic density data to mutable Dendritic Spine ?

mgeplf:
> Which method (API) is foreseen to add a post synaptic density data to mutable Dendritic Spine ?

`postSynapticDensity` has a non-const version, so you can get the vector that way

jacquemi-bbp:
Thank you.
What about how to add a root_section ? See my comment above.

mgeplf:
> append_root_section python method is not available for morphio.mut.DendriticSpine object

Go ahead and add it, and whatever other ones you need.

include/morphio/properties.h
```diff
@@ -158,9 +176,16 @@ struct CellLevel {
     bool diff(const CellLevel& other, LogLevel logLevel) const;
     bool operator==(const CellLevel& other) const;
     bool operator!=(const CellLevel& other) const;
-    std::string fileFormat() const;
-    uint32_t majorVersion();
-    uint32_t minorVersion();
+
+    std::string fileFormat() const {
+        return std::get<0>(_version);
```
jacquemi-bbp, line 181: This change does not pass the compilation step with clang 
https://github.com/BlueBrain/MorphIO/runs/2718243922#step:5:240

tests/test_6_writers.py
```diff
@@ -72,7 +72,7 @@ def test_write_basic():
         assert_array_equal(ImmutMorphology(Path(tmp_folder, "test_write.swc")).points, expected)
         h5_morph = ImmutMorphology(Path(tmp_folder, "test_write.h5"))
         assert_array_equal(h5_morph.points, expected)
-        assert h5_morph.version == ('h5', 1, 2)
+        assert h5_morph.version == ('h5', 1, 3)
```
mgeplf, line 75: don't worry about changing these tests; once you're happy w/ the API, I'll do the final pass to on tests and such.

jacquemi-bbp:
Saving a DendriticSpine instance with an empty post_synaptic_density list create a non readable file.

```python
spine_morph = morphio.DendriticSpine(file_path)

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-3-e2ba7bc60461> in <module>
----> 1 spine_morph = morphio.DendriticSpine(file_path)

RuntimeError: Unable to open the group "organelles/postsynaptic_density": (Symbol table) Object not found
```

jacquemi-bbp:
At the runtime, mut.DendriticSpine base class is not morphio._morphio.mut.Morphology as it is the case for mut.GlialCell

Is there a reason for that or could we modify the code do get the same parent class, morphio._morphio.mut.Morphology ? 

```python
glial = morphio.mut.GlialCell()
glial.__class__.__bases__
>> (morphio._morphio.mut.Morphology,)

spine=morphio.mut.DendriticSpine()
spine.__class__.__bases__
>> (pybind11_builtins.pybind11_object,)
```

jacquemi-bbp:
I tested the API, it works for creating, writing and reading DendriticSpine.
It would be nice to correct the mut.DendriticSpine base class

mgeplf:
Cool, I'll make that fix, and espcially the `Saving a DendriticSpine instance with an empty post_synaptic_density list creates a non readable file.` from above.

Note: any files created with this WIP will have to be deleted, as the section types are wrong.

include/morphio/enums.h
mgeplf: @jacquemi-bbp thinking about the numbering here, I wonder if it makes sense to keep the 'work around'; does the `head` behave like a `soma`?  Is there only a single one of them, for instance?  In the documentation, they seem to be more common, and actually attach to different necks

jacquemi-bbp: There can be several heads, attached to the same "neck" section.

mgeplf: Yeah, that's why I think that it's not really comparable to a `soma` and why using a different constant makes sense.

jacquemi-bbp: Furthermore, I made a mistake on the documentation example. The green section is neck type (2) and the red one is head type (1) as the blue one

mgeplf: Could you update the documentation with the new IDs, and also the changes needed to fix the mistake?

jacquemi-bbp: https://bbpcode.epfl.ch/code/#/c/53854/

include/morphio/dendritic_spine.h
```diff
@@ -0,0 +1,32 @@
...
+    const std::vector<Property::DendriticSpine::PostSynapticDensity>& postSynapticDensity() const
+        noexcept;
+
+  protected:
+    friend class mut::DendriticSpine;
+    DendriticSpine(const Property::Properties& properties);
+
+  private:
+    Soma soma() const;
+    Mitochondria mitochondria() const;
```
jacquemi-bbp, line 25: From my understanding, there is no Mitochondria and no Endoplasmic Reticulum object attached to Dendritic Spine.

include/morphio/enums.h
jacquemi-bbp, line 66: From last morphology documenttion commit here:
https://bbpcode.epfl.ch/code/#/c/53854/2/source/h5v1.rst

 For dendritic spine the values are: 2: `neck`, 3: `head`.

jacquemi-bbp (changes requested):

include/morphio/dendritic_spine.h
mgeplf: Yeah, that's why they are marked private so if anyone tries to access them, they won't be able to.

include/morphio/enums.h
mgeplf: Will fix.

jacquemi-bbp (approved):

Issue #324, opened 2021-06-04
Markers are only parsed if at the top/botom of file


arnaudon:
But not if in the middle, as in this example:

```
(FilledCircle
    (Color Red)
    (Name "Marker detected 1")
    (  451.82  -244.42   -20.80     0.09)  ; 1
  )  ;  End of markers

("CellBody"                                     
  (Color Green)
  (CellBody)                                    
  (  435.99  -231.01   -20.29     0.24)  ;  1, 1
)  ;  End of contour

( (Color RGB (0, 221, 0))                                             
  (Axon)
  (  453.15  -243.36   -21.06     3.30)  ; Root                       
  (  454.26  -242.59   -19.77     2.89)  ; 1, R                       
                                                                      
  (FilledCircle                                                       
    (Color RGB (0, 221, 0))                                           
    (Name "Marker undetected 1")                                      
    (  453.05  -243.48   -20.80     0.09)  ; 1                        
  )  ;  End of markers                                                
   Incomplete                                                         
)  ;  End of tree                                                     
                                                                      
( (Color Red)                                                         
  (Dendrite)                                                          
  (  450.09  -243.77   -21.14     3.72)  ; Root                       
  (  451.99  -244.35   -21.18     3.72)  ; 1, R                       
                                                                      
  (FilledCircle                                                       
    (Color Red)                                                       
    (Name "Marker undetected 2")                                      
    (  451.82  -244.42   -20.80     0.09)  ; 1                        
  )  ;  End of markers                                                
  (                                                                   
    (  469.69  -257.10   -22.37     1.17)  ; 1, R-1                   
    (  470.64  -258.24   -22.32     0.87)  ; 2                        
  ) 
) 
(FilledCircle
    (Color Red)
    (Name "Marker detected 2")
    (  451.82  -244.42   -20.80     0.09)  ; 1
  )  ;  End of markers
```
Would it be possible to detect them as well, so that  `morphio.Morphology('morph.asc').markers` has them all? 

Thank you!

asanin-epfl:
Do you need it for all types of markers or only for "Incomplete" markers?

arnaudon:
Ah sorry, this one is not about Incomplete, but 'normal markers'. I named the 2 markers which are not detected by morphio as `undetected`. The incomplete marker is in the other issue.

asanin-epfl:
[...]

PR #325, opened 2021-06-09
Parse markers at any level of ASC morphology

asanin-epfl:
Solution to #324


asanin-epfl:
@tomdele can you please look why MacOS tests can't setup? There is a problem with brew.

mgeplf:
> @tomdele can you please look why MacOS tests can't setup? There is a problem with brew.

I have a fix for it in libsonata, haven't had time to add it here; I'll do it now.

asanin-epfl:
> I have a fix for it in libsonata, haven't had time to add it here; I'll do it now.

ok, I forgot to delete debug prints in the test file, so please don't merge the PR.

arnaudon:
thank you!

src/readers/morphologyASC.cpp
```diff
@@ -287,10 +287,19 @@ class NeurolucidaParser
                     lex_.consume();
                 }
                 lex_.consume(Token::RSPINE, "Must be end of spine");
+            } else if (skip_sexp(+id)) {
+                lex_.consume_until_balanced_paren();
             } else if (id == Token::LPAREN) {
                 if (skip_sexp(peek_id)) {
                     // skip words/strings/markers
```
mgeplf, line 294: markers are no longer included

tomdele (approved):
looks good to me !

src/readers/morphologyASC.cpp
```diff
@@ -287,10 +287,19 @@ class NeurolucidaParser
                     lex_.consume();
                 }
                 lex_.consume(Token::RSPINE, "Must be end of spine");
+            } else if (skip_sexp(+id)) {
```
mgeplf, line 290: why do you need this? I'm worried that this will skip something we shouldn't be skipping (ie: that doens't come right after a `LPAREN`.

mgeplf (commented):
was this tested against our complete corpus of asc files?

src/readers/morphologyASC.cpp
asanin-epfl: That's a good point. Thank you. I've rebuilt it.

src/readers/morphologyASC.cpp
asanin-epfl: done

asanin-epfl:
> need to see if this at least parses everything we have

@mgeplf I've updated the parser so now all the changes are located within the dedicated MARKER case. Also I tested this PR against MorphologyRepository that you gave me. It passed fine. Also it showed that there are hidden corrupted morph files in MorphologyRepository. Here is the test:

```python
def test_against_morphology_repository():
    def test(path: Path):
        for file in path.iterdir():
            # there are hidden corrupted files in this repo, we want to skip such files
            if file.is_file() and file.suffix.lower() == '.asc' and not file.name.startswith('.'):
                m = Morphology(file)
                assert len(m.sections) > 0, f'{file} has no sections'
                assert len(m.root_sections) > 0, f'{file} has no root sections'
            elif file.is_dir():
                test(file)

    repo_dir = Path('/Users/sanin/workspace/MorphologyRepository')
    test(repo_dir)
```

asanin-epfl:
Do python tests account for in coverage? The coverage report shows that the new added code lines of this PR are not covered. However I've added tests for markers in `test_2_neurolucida.py`.

tomdele:
No they don t account for the coverage. Only the C++ side has the coverage set up. I am actually unsure if a tool exist to compute the coverage of a library from its bindings. It looks a bit specific and complex to do this ...

asanin-epfl:
Thank you for clarifying! I've added the C++ test.

asanin-epfl:
I've also added `parent_id` property to `Marker`. Let me know if I should revert it.

mgeplf:
> I've also added parent_id property to Marker. Let me know if I should revert it.

At first blush, I think this is a great idea; I will think about it further, as I review - which I will hopefully get to by Friday afternoon...please don't merge it until then.  Thanks for the PR.

include/morphio/properties.h
```diff
@@ -146,6 +146,7 @@ struct Annotation {
 struct Marker {
     PointLevel _pointLevel;
     std::string _label;
+    int32_t _parentId;  // id of parent section
```
mgeplf, line 149: so, classic question: is this 0 the soma, or not?  We need to be careful about this due to the soma not being considered a section...

asanin-epfl: Please elaborate here. `_parentId` can be -1 for cases when a marker belongs to the soma or marker belongs to another marker. See https://github.com/BlueBrain/MorphIO/blob/master/src/readers/morphologyASC.cpp#L146 and https://github.com/BlueBrain/MorphIO/blob/master/src/readers/morphologyASC.cpp#L141

mgeplf: I think that's offset from what is returned from `morphology::section`; section_id == 0 is a neurite, not the soma.

mgeplf: > _parentId can be -1 for cases when a marker belongs to the soma

Ah, ok, pardon me, that seems right.

mgeplf:
> Also I tested this PR against MorphologyRepository that you gave me. It passed fine.

Thanks, I also tried converting all files, and they match between MorphIO 3.2 and this dev version.

Nice work.

mgeplf (approved):

asanin-epfl:
Thank you!

asanin-epfl:
@mgeplf btw is `parent_id` name ok? Should I rather name it `section_id`?

mgeplf:
> btw is parent_id name ok? Should I rather name it section_id?

Hmm, good point.  Now that you mention it, `section_id` gives more info on how it could be used - however, if we also use it for the `Incomplete` markers, does it make sense?  If we call it `parent_id`, we can specify in the docs what it means for various different markers, but if we use `section_id`, it's more definitive.

asanin-epfl:
> if we also use it for the `Incomplete` markers, does it make sense?

Yes because `Incomplete` marks the section as incomplete, e.g. a section with a cut leaf.

mgeplf:
Ok; I'm fine with either - up to you if you want to change it while we have the chance.

asanin-epfl:
Thank you! The PR is ready #329

Issue #323, opened 2021-06-04
Parsing  of 'Incomplete' markers


arnaudon:
It could be nice to somehow retrieve the 'Incomplete' markers from morphio python API, such as this one:

https://github.com/BlueBrain/MorphIO/blob/5a71472f425a1cb336212dfe871cadabaf1a84af/tests/test_2_neurolucida.py#L486

asanin-epfl:
`Incomplete` are not markers according to the source files of [morphologyASC.cpp](https://github.com/BlueBrain/MorphIO/blob/master/src/readers/morphologyASC.cpp). It has to be decided how to represent 'Incomplete' as they don't have points as usual markers. For example there is an 'Incomplete'. How to represent it on the morphology? You would probably need to understand where it is located within the morphology.

arnaudon:
You mean this: https://github.com/BlueBrain/MorphIO/blob/master/src/readers/morphologyASC.cpp#L33
and this? https://github.com/BlueBrain/MorphIO/blob/master/src/readers/morphologyASC.cpp#L49
The incomplete is not a marker indeed, it's a 'tag' on leave sections. 

From my basic knowledge of morphio, I can see something like `section.is_complete`, similar to `section.is_root`, which is True by default, unless the section is followed by the `Incomplete` tag in the file?

asanin-epfl:
> You mean this: ...

Rather this https://github.com/BlueBrain/MorphIO/blob/master/src/readers/lex.cpp#L202 but that's not relevant. 'Incomplete' is not a marker.

> From my basic knowledge of morphio, I can see something like section.is_complete, similar to section.is_root, which is True by default, unless the section is followed by the Incomplete tag in the file?

That's interesting. However HDF5 and SWC files don't have 'Incomplete'. For them it would always be empty. What's the point of it then? If there is a property `incomplete` on the morphology instance then what would you want to see inside this property?

arnaudon:
Yes, this is specific to ascii files generated by neuroluscida, so if it cannot be made optional (like if .asc with such tags then we have .is_complete), then set it to True by default (no info means it is complete). 

What we would need is the list of incomplete leave sections from these tags, one way or another. Maybe `morph.incomplete_sections = [sec_id1, sec_id2, ...]` ? Just go for the simplest option, we can change the format as we wish later in python. 

[...]

PR #328, opened 2021-06-14
Consider 'Incomplete' as a marker

asanin-epfl:
Solution to #323


arnaudon:
Wow, thanks for that!

src/readers/morphologyASC.cpp
```diff
@@ -281,6 +281,12 @@ class NeurolucidaParser
                 }
                 return true;
             } else if (is_end_of_branch(id)) {
```
tomdele, line 283: I am really bad with the asc format.  Are we sure the `Incomplete` is always detected as a end of branch ? We can never have something like : 

```
( -266.79  -131.77   -26.13     
Incomplete       
```
or even:
```
 ( -266.79  -131.77   -26.13     0.92) 
 Incomplete
 ( -268.17  -130.62   -24.75     0.92)
```
before a end of split ?

asanin-epfl: I understand that 'Incomplete' always ends the branch from my discussion with @arnaudon. We can additionally validate other cases and throw.

arnaudon: Yes, it is to mark a branch in incomplete, so it must be at a leave branch. I think you can safely throw if not

tomdele (approved):
Looks Great ! 
Thanks Aleksei

asanin-epfl:
Thank you!

PR #332, opened 2021-07-30
don't compile c++ tests when creating python package

mgeplf:
* thus we don't need a c++17 compiler


AurelienJaquier (approved):
Looks good to me

matz-e (approved):

PR #335, opened 2021-08-27
Change tmp dir

mgeplf:


tests/utils.py
```diff
@@ -33,16 +24,16 @@ def ignored_warning(warning):
     finally:
         set_ignored_warning(warning, False)
 
+
 @contextmanager
-def _tmp_file(content, extension):
-    with tempfile.NamedTemporaryFile(suffix='.' + extension, mode='w') as tmp_file:
+def _tmp_file(tmp_folder, content, extension):
+    suffix = '.' + extension
+    with tempfile.NamedTemporaryFile(suffix=suffix,
```
asanin-epfl, line 31: Can you please make a comment here why this logic is chosen? So no one tries to optimize this in the future.

mgeplf: which logic?

asanin-epfl: that this is required for Windows, no?

asanin-epfl: thank you

asanin-epfl (approved):

tests/utils.py
mgeplf: np, thanks for the review

PR #339, opened 2021-09-03
Remove unused python mock dependency

musicinmybrain:
While `mock` is in `tests/requirements_tests.txt`, nothing actually uses it. (Starting with Python 3.3, it is in the standard library as `unittest.mock` anyway.)


mgeplf:
Good eye, thanks.

PR #342, opened 2021-09-10
Mg/windows build

mgeplf:


setup.py
```diff
@@ -79,35 +49,34 @@ def build_extension(self, ext, cmake):
...
 
-with open('README.rst') as f:
+install_requires = ['numpy>=1.14.1',
+                    ]
+
+with open('README.rst', encoding='utf-8') as f:
     long_description = f.read()
 
+if platform.system() == 'Windows':
+    install_requires += ['h5py>=3,<4',   # use h5py's hdf5 install so we don't have to redistrute hdf5
```
mgeplf, line 70: redistrute fixed in next patch

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "wheel", "cmake", "ninja"]
+build-backend = "setuptools.build_meta"
```
matz-e, line 3: So we're doing this now?

setup.py
```diff
@@ -79,35 +49,34 @@ def build_extension(self, ext, cmake):
...
-                                           "Visual Studio 15 2017 Win64")]
-        else:
-            cmake_args += ['-DCMAKE_BUILD_TYPE={}'.format(cfg),
-                           '-DMorphIO_CXX_WARNINGS=OFF',
-                           ]
-            build_args += ["--", "-j{}".format(max(MIN_CPU_CORES, get_cpu_count())),
-                           ]
+        cmake_args += ['-DCMAKE_BUILD_TYPE={}'.format(cfg),
+                       '-DMorphIO_CXX_WARNINGS=OFF',
+                       '-GNinja',
```
matz-e, line 54: Neat (note to self: requires new dependencies in Spack)

matz-e (approved):
Looks nice, and I appreciate the more compact `setup.py`.

matz-e:
BTW, please change the title of this :)

setup.py
mgeplf: yeah, that's why I added you to the review; it should be ok though, right? adding `ninja`?

pyproject.toml
mgeplf: This is the way I got things to be 'reliable'; I hate that I don't know the full interactions of this; however, it seems that it doesn't work as well as I expected - I will go back to the drawing board.

setup.py
matz-e: `depends_on('ninja', type='build')` should probably be good enough.

matz-e:
Approval does not reset on new pushes?

mgeplf:
> Approval does not reset on new pushes?

apparently not

matz-e:
> > Approval does not reset on new pushes?
> 
> apparently not

Well, I'm still in approval

PR #343, opened 2021-09-14
cleanup building and publishing wheels

mgeplf:


.github/workflows/publish-sdist-wheels.yml
```diff
@@ -66,35 +69,41 @@ jobs:
...
       - name: Store sdist as artifact
         uses: actions/upload-artifact@v2
         with:
           name: dist
-          path: dist
+          path: dist/*.whl
 
   upload_wheels:
+    name: Upload wheels to PyPI
+    if: github.event_name == 'release' && github.event.action == 'published'
```
asanin-epfl, line 96: Am I correct that this is a replacement for
```yaml
on:
  push:
    tags:
      - 'v[0-9]+.[0-9]+.[0-9]+'
```
from before on the lines 3-6

asanin-epfl:
I am not sure how qualified I am to review here. Looks good. I can approve if necessary.

.github/workflows/publish-sdist-wheels.yml
mgeplf: yes, that is the plan - whether it works - I will find out when I try the release.

However, the reasoning is that I want to have the build artifacts to test before they are uploaded to pypi; and they should exist for each commit (FYI).

matz-e: Seems legit. Would not want to have wheel building break and only find out at release time.

matz-e (approved):
Looks legit.

.github/workflows/publish-sdist-wheels.yml
mgeplf: Yeah, that's is a nightmare that I luckily haven't had to live through - this should be safe now.

PR #344, opened 2021-09-15
Update setup

mgeplf:


matz-e (approved):

mgeplf:
I should note that I tried building this in a clean `ubuntu-20.04` with both the basic `pip` and the `pip --use-feature=in-tree-build`, and it seemed to have worked.

Issue #345, opened 2021-09-20
Failed building of wheels during building of documentation


asanin-epfl:
The full log can be seen at https://readthedocs.org/projects/morphio/builds/14720285/. It started 5 days ago. @mgeplf please advice.
```
Failed to build MorphIO

Pip subprocess error:
  DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
   pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
  ERROR: Command errored out with exit status 1:
   command: /home/docs/checkouts/readthedocs.org/user_builds/morphio/conda/latest/bin/python /home/docs/checkouts/readthedocs.org/user_builds/morphio/conda/latest/lib/python3.6/site-packages/pip/_vendor/pep517/in_process/_in_process.py build_wheel /tmp/tmpsnflxq8d
       cwd: /tmp/pip-req-build-li0b8iq4
 ...
  -- HDF5 C compiler wrapper is unable to compile a minimal HDF5 program.
  CMake Error at /tmp/pip-build-env-a_ep2bk0/overlay/lib/python3.6/site-packages/cmake/data/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
    Could NOT find HDF5 (missing: HDF5_LIBRARIES HDF5_INCLUDE_DIRS) (found
    version "")
  Call Stack (most recent call first):
    /tmp/pip-build-env-a_ep2bk0/overlay/lib/python3.6/site-packages/cmake/data/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
    /tmp/pip-build-env-a_ep2bk0/overlay/lib/python3.6/site-packages/cmake/data/share/cmake-3.21/Modules/FindHDF5.cmake:1001 (find_package_handle_standard_args)
    3rdparty/CMakeLists.txt:1 (find_package)
```

mgeplf:
My guess is that my changes to the build process don't work under conda - a use case I hadn't tried.  It looks like the `pyproject.toml` supplied cmake isn't finding the HDF5 install - I wonder if we could just delete the `pyproject.toml`; in the readthedocs build, and make sure that `cmake` & `ninja` are installed.

mgeplf:
[...]

PR #348, opened 2021-10-04
Work around conda cmake finding HDF5 issues by building docs with libhdf5-dev already installed

mgeplf:
* switching to pyproject.toml broke the read the docs build in conda
* it appears that the conda `cmake` finds `conda` installed packages
  fine, but if cmake is installed through `pyproject.toml`, that CMake
  doesn't find HDF5
* I don't want to change pyproject.toml building - it works in windows

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,5 +9,13 @@ submodules:
   include: all
   recursive: true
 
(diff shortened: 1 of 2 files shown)


mgeplf:
Fixes #345

GianlucaFicarelli (approved):

PR #351, opened 2021-10-06
dependencies: allow to use an external pybind11.

### LINKED ISSUE #337: Support external pybind11
In general, `pybind11` supports use with `add_subdirectory(extern/pybind11)` or from an install with `find_package(pybind11 CONFIG)` (https://pybind11.readthedocs.io/en/stable/cmake/index.html, https://pybind11.readthedocs.io/en/stable/compiling.html#building-with-cmake).

I’ve tried to do this downstream in Fedora Linux, but couldn’t manage to untangle CMake enough to get it working properly in this particular case. An `EXTERNAL_PYBIND11` option, similar to the existing `EXTERNAL_HIGHFIVE`, would be a big help!

This isn’t a critical issue, as Fedora Linux guidelines do allow us to continue to bundle `pybind11` as long as upstream does not support building with an external copy.

matz-e:
Fixes #337.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,8 +7,8 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 
 option(BUILD_BINDINGS "Build the python bindings" ON)
 option(MorphIO_CXX_WARNINGS "Compile C++ with warnings" ON)
-option(EXTERNAL_HIGHFIVE "Use HighFive from external source" ON)
-option(EXTERNAL_PYBIND11 "Use pybind11 from external source" ON)
+option(EXTERNAL_HIGHFIVE "Use HighFive from external source" OFF)
+option(EXTERNAL_PYBIND11 "Use pybind11 from external source" OFF)
 option(MORPHIO_TESTS "Build tests" ON)
 option(MORPHIO_USE_DOUBLE "Use doubles instead of floats" OFF)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,8 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 
 option(BUILD_BINDINGS "Build the python bindings" ON)
 option(MorphIO_CXX_WARNINGS "Compile C++ with warnings" ON)
-option(EXTERNAL_HIGHFIVE "Use HighFive from external source" OFF)
+option(EXTERNAL_HIGHFIVE "Use HighFive from external source" ON)
+option(EXTERNAL_PYBIND11 "Use pybind11 from external source" ON)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -30,6 +31,9 @@ set(CMAKE_CXX_FLAGS "${FLAGS}")
 if (EXTERNAL_HIGHFIVE)
   find_package(HighFive REQUIRED)
 endif()
+if (EXTERNAL_PYBIND11)
+  find_package(pybind11 REQUIRED CONFIG)
```
mgeplf, line 35: should EXTERNAL_HIGHFIVE also have `CONFIG`?

I'm tempted to redo this like we have it in libsonata; ie the `EXTLIB_FROM_SUBMODULES` and then everything must be supplied.  Thoughts?

matz-e: :shrug: it's CMake, so I guess consistency would be nice for `CONFIG`. This just skips attempting to find a `Findpybind11` (or a permutation of capitalization).

In general, being consistent across our projects would be nice. So I'd be in favor of a `EXTLIB_FROM_SUBMODULES`.

mgeplf: > This just skips attempting to find a Findpybind11 (or a permutation of capitalization).

My CMake-Fu isn't great - is the HighFive CMake similar enough to the pybind11 that it will work w/ `CONFIG` in all cases?

> So I'd be in favor of a EXTLIB_FROM_SUBMODULES.

The question then is whether `lexertl14` and `GSL_LITE` should be part of that; seems unlikely that other use those packages, and having `pybind11` and `HighFive` only behind `EXTLIB_FROM_SUBMODULES` seems weird.  I'm vacillating; maybe it's best to have what you proposed here.

matz-e: HighFive should also work with `CONFIG`, I don't think any sane person should have written `FindHighFive.cmake`.

`GSL_LITE` is actually the one that bugs me the most — because its headers get installed alongside MorphIO, which violates my sense of package integrity. That's why we tend to use the external HighFive for C++.

OTOH, one would have to add a Spack package for `gsl-lite`, which I'm currently not inclined to do. MorphIO is currently the only consumer, and I'd hope that we could move to `std::ranges` at some point soon™.

mgeplf: This has been sitting too long, I'm going to go with what you have here.

mgeplf: (sitting too long, waiting for me, to be clear)

mgeplf (approved):

mgeplf:
Thank you very much!  Sorry again for letting this linger so long.

PR #352, opened 2021-10-12
Add Vasculature section_offsets & section_connectivity

### LINKED ISSUE #159: Vasculature section connectivity
Similarly to the neuronal morphologies, the section connectivity is also useful for vasculature datasets.

```python
from morphio.vasculature import Vasculature
v = Vasculature(filepath)
v.connectivity
```

### LINKED ISSUE #160: Vasculature section offsets
Similarly to the neuronal morphologies, it is useful to access the section offsets for the vasculature.

eleftherioszisis:
Closes #160
Closes #159

### DIFF
--- a/binds/python/bind_vasculature.cpp
+++ b/binds/python/bind_vasculature.cpp
@@ -39,8 +39,9 @@ void bind_vasculature(py::module& m) {
 
         .def_property_readonly(
             "section_offsets",
-            [](morphio::vasculature::Vasculature& vasculature)
-                { return as_pyarray(vasculature.sectionOffsets()); },
+            [](morphio::vasculature::Vasculature& vasculature) {
+                return as_pyarray(vasculature.sectionOffsets());
+            },
             "Returns a list with offsets to access data of a specific section in the points\n"
             "and diameters arrays.\n"
             "\n"
--- a/binds/python/bind_vasculature.cpp
+++ b/binds/python/bind_vasculature.cpp
@@ -36,6 +36,20 @@ void bind_vasculature(py::module& m) {
              "throw RawDataError if the id is out of range",
              "section_id"_a)
 
+
+        .def_property_readonly(
+            "section_offsets",
+            [](morphio::vasculature::Vasculature& vasculature)
+                { return as_pyarray(vasculature.sectionOffsets()); },
+            "Returns a list with offsets to access data of a specific section in the points\n"
+            "and diameters arrays.\n"
+            "\n"
+            "Example: accessing diameters of n'th section will be located in the DIAMETERS\n"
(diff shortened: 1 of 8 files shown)


binds/python/bind_vasculature.cpp
```diff
@@ -36,6 +36,21 @@ void bind_vasculature(py::module& m) {
...
+            [](morphio::vasculature::Vasculature& vasculature) {
+                return as_pyarray(vasculature.sectionOffsets());
+            },
+            "Returns a list with offsets to access data of a specific section in the points\n"
+            "and diameters arrays.\n"
+            "\n"
+            "Example: accessing diameters of n'th section will be located in the DIAMETERS\n"
+            "array from DIAMETERS[sectionOffsets(n)] to DIAMETERS[sectionOffsets(n+1)-1]\n"
+            "\n"
+            "Note: for convenience, the last point of this array is the points array size\n"
```
mgeplf, line 51: not sure I follow; I think the word `point` in `the last point of this array` is confusing, but I'm still not following when I substitute `element`

tests/test_10_vasculature.py
```diff
@@ -63,6 +65,34 @@ def test_section_types():
...
+
+    offset = 0
+    expected_offsets = [offset]
+
+    for sec in morphology.sections:
+
+        offset += len(sec.points)
+        expected_offsets.append(offset)
+
+    npt.assert_allclose(morphology.section_offsets, expected_offsets)
```
mgeplf, line 79: can you add an example of using the `...for convenience,...` with a comment, please?

binds/python/bind_vasculature.cpp
eleftherioszisis: Let's say we have the following points (I list their ids):
`[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]`

And three sections:
```
S1: [0, 1, 2]
S2: [3, 4, 5]
S3: [6, 7, 8, 9]
```
The offsets will be:
`[0, 3, 6, 10]`

So the last entry is equal to the size of the points dataset. This allows for applying a diff for instance to calculate the section length (in terms of points) for each section without having to allocate an extra array to add that extra value.

mgeplf: ah, ok; basicaly if that example can be copied to the test in `tests/test_10_vasculature.py` that would be good

eleftherioszisis: sure, I copied it there.

tests/test_10_vasculature.py
eleftherioszisis: done

mgeplf (approved):

PR #353, opened 2021-10-18
ci: trigger a build of downstream morpho-kit

matz-e:
Should fullfil the requirements outlined in NSETM-915 by triggering a
build of morpho-kit whenever a push to the main branch of MorphIO
occurs.

Needs to have the corresponding token set, still.


matz-e:
Closing. I think using a webhook would probably be simpler.

PR #356, opened 2021-11-10
Check for supported python versions

eleftherioszisis:
A check is added in `setup.py` which results in a `RuntimeError` if morphio is pip installed in a python environment for which there are no wheels available.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,5 +97,5 @@ setup(
     setup_requires=[
         'setuptools_scm',
     ],
-    python_requires=">=3.7"
+    python_requires=">=3.7",
 )
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,10 @@ from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext
 
 
+if sys.version_info[:2] < (3, 7):
+    raise RuntimeError("Python version >= 3.7 required.")
+
+
 class CMakeExtension(Extension):
     def __init__(self, name, sourcedir=''):
         Extension.__init__(self, name, sources=[])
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -8,6 +8,10 @@
 from setuptools.command.build_ext import build_ext
 
 
+if sys.version_info[:2] < (3, 7):
+    raise RuntimeError("Python version >= 3.7 required.")
```
GianlucaFicarelli, line 12: As discussed with @eleftherioszisis , we could use `python_requires` as described in https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

It requires `pip>=9.0.0` and `setuptools>=24.2.0`, but these dependencies are pretty old (2016) and already required (or at least strongly recommended) to use the new dependency resolution algorithm

mgeplf: I agree with this, in general.

Do we really need to clamp the version? we just don't support 3.6, but ppl can build it if they want, no?

eleftherioszisis: The issue is that people attempt to pip install morphio in python3.6 environments and it breaks their installation without a clear message. Alternatively, we could emit a warning instead of making it this strict.

mgeplf: yeah; since 3.6 is so close to deprecation, let's do the same as `bluepysnap`:
https://github.com/BlueBrain/snap/pull/150/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R43

setup.py
```diff
@@ -96,5 +96,6 @@ def build_extension(self, ext):
     use_scm_version=True,
     setup_requires=[
         'setuptools_scm',
-        ],
+    ],
+    python_requires=">=3.7"
```
mgeplf, line 100: missing a trailing `,`

GianlucaFicarelli (approved):

PR #359, opened 2021-11-29
Add `n_points` methods for immutable Morphology & Vasculature

### LINKED ISSUE #357: Inexpensive number of points property for Section and Morphology
To check for the number of points of a section in python, we usually do:
```python
n_points = len(section.points)
```
However, that requires creating the points dataset on the python side so that its length is determined.

I would like to introduce an option where the number of points can be determined inexpensively using a property or the `__len__` dunder.
Possible variations:
```python
n_points = section.n_points
n_points = section.number_of_points
n_points = len(section)
```
And the same would be applied to the entire Morphology so that we don't have to copy all its points to find their number.

eleftherioszisis:
Closes #357

### DIFF
--- a/binds/python/bind_immutable.cpp
+++ b/binds/python/bind_immutable.cpp
@@ -87,6 +87,12 @@ void bind_immutable_module(py::module& m) {
             "Returns a list with all points from all sections (soma points are not included)\n"
             "Note: points belonging to the n'th section are located at indices:\n"
             "[Morphology.sectionOffsets(n), Morphology.sectionOffsets(n+1)[")
+
+        .def_property_readonly(
+            "n_points",
+            [](const morphio::Morphology& obj) { return obj.points().size(); },
+            "Returns the number of points from all sections (soma points are not included)")
+
         .def_property_readonly(
             "diameters",
             [](const morphio::Morphology& morpho) {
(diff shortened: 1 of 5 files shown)


adrien-berchet (commented):
Looks good to me, thanks!
I am just wondering if the property of the `Morphology` should not be called `n_section_points` to be more clear?

eleftherioszisis:
> Looks good to me, thanks! I am just wondering if the property of the `Morphology` should not be called `n_section_points` to be more clear?

Given that there are no other points to be confused within the morphology, isn't it redundant?

adrien-berchet:
> > Looks good to me, thanks! I am just wondering if the property of the `Morphology` should not be called `n_section_points` to be more clear?
> 
> Given that there are no other points to be confused within the morphology, isn't it redundant?

It's just to make more explicit that the soma points are not counted. But as you prefer, both are good to me, it was just a comment.

eleftherioszisis:
The main reason I prefer `n_points` is that it reflects the `points` dataset. Introducing a name with a different infix would make the user think that it might be something more than a counting of the `points` dataset.

adrien-berchet:
Yeah that's why both are good to me, I don't know if one is better than the other for a new user but let's approve like this :)

adrien-berchet (approved):

mgeplf:
I think the point about the naming is important; instead of `n_section_points`, would `n_neurite_points` work?

Depending on what underlying file ppl are using, they may consider `points` and `sections` to include the soma...

eleftherioszisis:
> I think the point about the naming is important; instead of `n_section_points`, would `n_neurite_points` work?
> Depending on what underlying file ppl are using, they may consider `points` and `sections` to include the soma...

morphio has no concept of neurites as it works exclusively with sections. Therefore, if a different name is to be used, `n_section_points` would be more suitable imho. However, as I said above I like how `n_points` maps intuitively to the `points` dataset, because otherwise `points` should be have beed renamed into `section_points`.

Should I use `n_section_points` as @adrien-berchet suggested?

mgeplf:
> [...] I like how n_points maps intuitively to the points dataset, because otherwise points should be have beed renamed into section_points.

Ok, I see the sense in it, your initial intuition was spot on.

mgeplf (approved):

PR #360, opened 2021-12-08
is-heterogeneous

mgeplf:


src/mut/section.cpp
```diff
@@ -59,6 +60,16 @@ bool Section::isRoot() const {
     return true;
 }
 
+bool Section::is_heterogeneous(bool downstream) const {
+    auto p = [&](const std::shared_ptr<Section>& s) { return type() != s->type(); };
```
mgeplf, line 64:
```suggestion
    auto predicate = [&](const std::shared_ptr<Section>& s) { return type() != s->type(); };
```

I should be more verbose when prototyping things; good call on using `type()`

mgeplf (commented):
couple nits, otherwise LGTM

eleftherioszisis (approved):

PR #365, opened 2021-12-17
Update readthedocs python to 3.9

eleftherioszisis:
Dropping the wheels for python3.6 without changing the python version in the docs environment resulted into not having wheels available and therefore trying to compile it and exceeding the max available memory.

Updating the python version results in using the wheels and therefore building the documentation without an issue:
https://readthedocs.org/projects/morphio/builds/15565753/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -27,3 +27,5 @@ python:
   install:
     - method: pip
       path: .
+      extra_requirements:
+        - docs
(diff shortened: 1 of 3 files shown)


mgeplf (approved):

eleftherioszisis:
In the end, conda was replaced by pip, because it was the reason for allocating so much memory for building the package from source.

PR #366, opened 2021-12-17
Remove Morphology friend from Soma

eleftherioszisis:
To remove the `Morphology` friend, I had to make the `Soma` constructor public. Would that be undersirable?


include/morphio/soma.h
```diff
@@ -65,15 +65,10 @@ class Soma
      */
     floatType maxDistance() const;
 
-  private:
     explicit Soma(const std::shared_ptr<Property::Properties>&);
```
mgeplf, line 68: I believe that it was done this way b/c the public interface is smaller; which means less potentially breaking changes and ppl depending on certain implementation details. (ie: Hyrum's Law); that's why I'd lean to not make it public.

eleftherioszisis: However, resorting to using friends to bypass the private constructor seems like an unnecessary complexity to me, isn't it?

include/morphio/soma.h
```diff
@@ -25,22 +26,33 @@ namespace morphio {
...
+    Soma() = default;
+    Soma(const Soma& soma) = default;
+
+    explicit Soma(const Property::Properties& properties);
+    explicit Soma(const Property::PointLevel& point_properties);
+
     /// Return the  coordinates (x,y,z) of all soma points
-    range<const Point> points() const noexcept {
-        return properties_->_somaLevel._points;
+    std::vector<Point>& points() noexcept {
```
matz-e, line 38: So… this makes things mutable, is this still the right place then?

include/morphio/soma.h
```diff
@@ -25,22 +26,33 @@ namespace morphio {
...
+        return properties_._points;
+    }
+    const std::vector<Point>& points() const noexcept {
+        return properties_._points;
     }
 
     /// Return the diameters of all soma points
-    range<const floatType> diameters() const noexcept {
-        return properties_->_somaLevel._diameters;
+    std::vector<morphio::floatType>& diameters() noexcept {
```
matz-e, line 46: Same here.

include/morphio/soma.h
eleftherioszisis: Given that there is no readonly/mutable duality with these changes, I thought I might leave it in `morphio/soma.h`. But you are right, it can be moved to mut too.

matz-e: I _suppose_ one could say that the Morphology proper is the (im)mutable entity, in which case the placement of the soma would be fine. But then makes me wonder, shouldn't the section class receive the same treatment?

eleftherioszisis: Well, this is an experiment to check if I can pull this through. pybind11 is giving me hell at the time being...
If it will work, yes, I don't see why not do the same for sections and reduce the complexity of the code.

eleftherioszisis: Although with sections, the switching from leveraging the ranges to copying the data so that a section data structure is created the same way for both mut/readonly, will result in quite a performance penalty. With soma it doesn't really matter given that it is just one and with very few points.

src/soma.cpp
```diff
@@ -18,11 +21,20 @@ Soma::Soma(const Property::PointLevel& point_properties)
 
 
 Point Soma::center() const {
+
+    if (points().empty()) {
+        return {NaN, NaN, NaN};
```
matz-e: Do people check for NaN? If I use Python, I would just expect this to return `None`, and by now, in compiled languages, expect `std::optional` (or language equivalent)

eleftherioszisis: yeah, I removed them after thinking twice and also realizing that the volume/surface functions already raise exceptions.

PR #368, opened 2022-01-17
Cleanup: The Sequel

eleftherioszisis:


eleftherioszisis:
Thanks for the test @mgeplf !

src/vasc/section.cpp
```diff
@@ -7,78 +7,78 @@ namespace vasculature {
...
 
 uint32_t Section::id() const noexcept {
-    return _id;
+    return id_;
 }
 
 template <typename TProperty>
 range<const typename TProperty::Type> Section::get() const {
-    auto& data = _properties->get<TProperty>();
+    const auto& data = properties_->get<TProperty>();
```
mgeplf, line 55: nice

src/vasc/section.cpp
```diff
@@ -7,78 +7,78 @@ namespace vasculature {
...
 std::vector<Section> Section::predecessors() const {
     std::vector<Section> result;
-    const auto it = _properties->predecessors().find(_id);
-    if (it != _properties->predecessors().end()) {
+    const auto it = properties_->predecessors().find(id_);
+    if (it != properties_->predecessors().end()) {
         result.reserve(it->second.size());
-        for (const uint32_t id_ : it->second) {
-            result.emplace_back(id_, _properties);
+        for (uint32_t id : it->second) {
```
mgeplf, line 68: also nice

src/vasc/vasculature.cpp
```diff
@@ -36,28 +36,28 @@ Vasculature::Vasculature(const std::string& source) {
...
 }
 
 std::vector<Section> Vasculature::sections() const {
     std::vector<Section> sections_;
-    const auto& vasc_sections = _properties->get<property::VascSection>();
+    const auto& vasc_sections = properties_->get<property::VascSection>();
     sections_.reserve(vasc_sections.size());
-    for (size_t i = 0; i < vasc_sections.size(); ++i) {
-        sections_.emplace_back(static_cast<uint32_t>(i), _properties);
+    for (uint32_t i = 0; i < vasc_sections.size(); ++i) {
```
mgeplf, line 52: should we worry about very large amount of `vasc_sections`?

mgeplf (approved):
thanks, style uniformity is extremely helpful, IMO; thanks for suffering through that

src/vasc/vasculature.cpp
eleftherioszisis: If we are going to support whole brain datasets, I think we should yeah. Should I replace with size_t or uint64_t ?

mgeplf: `size_t` seems most natural to me

PR #370, opened 2022-01-21
Add section type exception for unsupported type in ascii writer

### LINKED ISSUE #369: saving custum types in .asc
We just bumped into this obscure error if we do the following (from neurom):

```
m = nm.load_morphology('test.asc')
for section in m.neurites[0].iter_sections():
    section.morphio_section.type = SectionType.custom5
m.write('test_custom.asc')
```

we get this crash:

```
----> 1 m.write("t.asc")

IndexError: map::at
```

Saving to .swc works, I guess in .asc one has to map the custom types to a header name of the section list, or something like that? 

It's not at all urgent at all, it's just to let you know!

eleftherioszisis:
Fixes #369

### DIFF
--- a/src/mut/writers.cpp
+++ b/src/mut/writers.cpp
@@ -154,7 +154,14 @@ static void _write_asc_points(std::ofstream& myfile,
 static void _write_asc_section(std::ofstream& myfile,
                                const Morphology& morpho,
                                const std::shared_ptr<Section>& section,
+                               const std::map<morphio::SectionType, std::string>& header,
                                size_t indentLevel) {
+
+    // allowed types are only the ones available in the header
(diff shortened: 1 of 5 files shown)


src/mut/writers.cpp
```diff
@@ -202,8 +208,13 @@ void asc(const Morphology& morphology, const std::string& filename) {
     }
 
     for (const auto& section : morphology.rootSections()) {
+        // allowed types are only the ones available in the header
```
mgeplf, line 211: looks good to me, could we just add a named type here, or something slightly clearer?

eleftherioszisis: yes, ofc!

mgeplf (approved):

PR #379, opened 2022-02-11
Have Section and MitoSection have value semantics for comparison

mgeplf:
* two sections are the same when they have the same
  points/diameters/etc, not if they belong to the same instance of the
  same morphology

### DIFF
--- a/include/morphio/section_base.h
+++ b/include/morphio/section_base.h
@@ -46,7 +46,7 @@ class SectionBase
     std::vector<T> children() const;
 
     /** Return the ID of this section. */
-    inline uint32_t id() const noexcept;
+    uint32_t id() const noexcept { return id_; }
 
   protected:
     SectionBase(uint32_t id, const std::shared_ptr<Property::Properties>& properties);
@@ -58,21 +58,6 @@ class SectionBase
     SectionRange range_;
     std::shared_ptr<Property::Properties> properties_;
 };
-
-template <typename T>
-inline bool SectionBase<T>::operator==(const SectionBase& other) const noexcept {
-    return other.id_ == id_ && other.properties_ == properties_;
-}
-
-template <typename T>
-inline bool SectionBase<T>::operator!=(const SectionBase& other) const noexcept {
-    return !(*this == other);
-}
-
-template <typename T>
-inline uint32_t SectionBase<T>::id() const noexcept {
-    return id_;
-}
 
 template <typename T>
 SectionBase<T>::SectionBase(uint32_t id, const std::shared_ptr<Property::Properties>& properties)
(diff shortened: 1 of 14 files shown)


mgeplf:
@eleftherioszisis: do you think that value semantics are what is expected for `==`; or should I have another method for comparison: `has_same_structure` or something?

eleftherioszisis:
I think the way you have it is fine. Shouldn't we also check that the topology is identical too, i.e. that they have the same parent/children?

eleftherioszisis:
Maybe have the convention that `==` would require everything to be identical, topology included, while `has_same_values` only the values?

mgeplf:
> `topology included`

I'm not sure that the number of children is intrinsic to the 'sameness' of the section; if it is, then would it need to be recursively true (ie: that the children are the same?) or just that the number matches?

I don't have a clear feeling of which one is 'right', if one can even say there is a 'right'.

eleftherioszisis:
Hmm, I was thinking in terms of ids, but I think you are right. If we include the topology we will more all less need to traverse the tree to ensure that the children and the parent are identical too. I can see two options:
1. delete the equality operator so that it cannot be used and leave the `has_same_values` method.
2. Limit the equality operator to check the ids of parent and children. I believe it is not very likely there will be two sections with the same id and differing values, right?

mgeplf:
I think I'll rename it `has_same_values`, and remove the equality (like you said).

include/morphio/section_base.h
```diff
@@ -98,6 +83,16 @@ SectionBase<T>::SectionBase(uint32_t id, const std::shared_ptr<Property::Propert
...
+template <typename T>
+inline bool SectionBase<T>::operator==(const SectionBase& other) const noexcept {
+    return other.id_ == id_ && other.properties_ == properties_;
+}
+
+template <typename T>
+inline bool SectionBase<T>::operator!=(const SectionBase& other) const noexcept {
+    return !(*this == other);
+}
+
```
eleftherioszisis, line 95: Just for our sanity, wouldn't it be better to also declare them on the binds side?
```
pybind11_class.def(pybind11::self == pybind11::self);
pybind11_class.def(pybind11::self != pybind11::self);
```
For both mut and immut?

src/section.cpp
```diff
@@ -55,6 +55,11 @@ range<const floatType> Section::perimeters() const {
     return get<Property::Perimeter>();
 }
 
+bool Section::hasSameShape(const Section& other) const noexcept {
```
eleftherioszisis, line 58: `hasSameShape` is a bit arbitrary (it usually refers to dimensions), isn't it? Wouldn't a name such as `hasSameData` or `hasSameValues` be a little more informative?

mgeplf: `Shape` I think is closest to what I mean; `Data`/`Value` is too generic; the diameters, etc, all define the shape of the section, but the child count doesn, imo.

include/morphio/section_base.h
mgeplf: python is already kinda doing this, from what I can tell.  If we need it, let's do it in another PR

eleftherioszisis: okk

src/section.cpp
eleftherioszisis: `hasSameGeometry` maybe?

eleftherioszisis (approved):

PR #380, opened 2022-02-11
update setup.py/pyproject.toml so the right numpy version is used

mgeplf:


eleftherioszisis (approved):

eleftherioszisis (commented):
It seems that pybind does not require numpy headers for compiling the lib. That means numpy is not needed at build time, isn't it?
https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#arrays

> Note that pybind11/numpy.h does not depend on the NumPy headers, and thus can be used without declaring a build-time dependency on NumPy; NumPy>=1.7.0 is a runtime dependency.

mgeplf:
> That means numpy is not needed at build time, isn't it?

Yes, that does seem to be the case; I should have noticed that when I was making these changes.   Thanks for noticing.

PR #386, opened 2022-02-18
Catch2: move to submodule and update to v2.13.8

matz-e:
Also add another `EXTERNAL_CATCH2` option to be able to use system
installations.


mgeplf (approved):
Thanks!

PR #388, opened 2022-03-01
Fix windows-2022 wheels & MANIFEST.in warnings

eleftherioszisis:
Warning example:
```warning: manifest_maker: MANIFEST.in, line 6: path '3rdparty/GSL_LITE/cmake/' cannot end with '/'```


matz-e (approved):

mgeplf (approved):

mgeplf:
For reference, I tested this with a windows 10 install, and all that was required was the `vc_redist.x64.exe` (the `Visual C++ 2015 Redistributable Package 64`); so the switch from the `v140` -> `v141` seems safe.

Thanks!

PR #389, opened 2022-03-01
don't use exceptions when non-exceptional methods are available

mgeplf:


include/morphio/section_base.h
```diff
@@ -124,19 +124,25 @@ T SectionBase<T>::parent() const {
 
 template <typename T>
 std::vector<T> SectionBase<T>::children() const {
+    const auto section_children = properties_->children<typename T::SectionId>();
```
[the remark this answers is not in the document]
mgeplf: thanks

matz-e (approved):

src/morphology.cpp
```diff
@@ -163,67 +227,4 @@ breadth_iterator Morphology::breadth_end() const {
     return breadth_iterator();
 }
 
-SomaType getSomaType(long unsigned int num_soma_points) {
-    try {
```
mgeplf, line 167: moved to anonymous namespace and replaced the `try` w/ a `switch`

src/morphology.cpp
```diff
@@ -13,10 +13,72 @@
...
+
+morphio::SomaType getSomaType(long unsigned int num_soma_points) {
+    switch (num_soma_points) {
+    case 0:
+        return morphio::SOMA_UNDEFINED;
+    case 1:
+        return morphio::SOMA_SINGLE_POINT;
+    case 2:
+        return morphio::SOMA_UNDEFINED;
+    default:;
```
mgeplf, line 48: this is how `clang-format` wanted this; FYI

src/morphology.cpp
```diff
@@ -13,10 +13,73 @@
...
+
+    std::string extension = source.substr(pos);
+
+    if (extension == ".h5" || extension == ".H5") {
+        return morphio::readers::h5::load(source);
+    } else if (extension == ".asc" || extension == ".ASC") {
+        return morphio::readers::asc::load(source, options);
+    } else if (extension == ".swc" || extension == ".SWC") {
+        return morphio::readers::swc::load(source, options);
+    }
```
eleftherioszisis, line 74: `else if` is not needed here because of the returns.

include/morphio/errorMessages.h
```diff
@@ -48,11 +48,11 @@ struct DebugInfo {
...
     /** Get section's line number within morphology file */
     int32_t getLineNumber(uint32_t sectionId) const {
-        try {
-            return _lineNumbers.at(sectionId);
-        } catch (const std::out_of_range&) {
+        const auto it = _lineNumbers.find(sectionId);
+        if (it == _lineNumbers.end()) {
             return -1;
         }
+        return it->second;
```
eleftherioszisis, line 55: This could also be:
```
return it == _lineNumbers.end() ? -1 : it->second;
```

mgeplf: I tried to avoid the `?:`; it's too easy to gloss over.

src/morphology.cpp
mgeplf: I'm going to leave it how it was.

eleftherioszisis (approved):

PR #392, opened 2022-03-07
HDF5 morphologies: rebuild only when options are passed

matz-e:
Merge the two `if` conditions to avoid additonal work.  This brings the reading time down by a factor of ~4-5 when reading a lot of data:
```
Morphologies		Time [ms] before	Time [ms] after
10			10			2
100			95			23
1000			1017			233
10000			11203			2318
```

### DIFF
--- a/src/morphology.cpp
+++ b/src/morphology.cpp
@@ -28,11 +28,9 @@ Morphology::Morphology(const Property::Properties& properties, unsigned int opti
 
     // For SWC and ASC, sanitization and modifier application are already taken care of by
     // their respective loaders
-    if (properties._cellLevel.fileFormat() == "h5") {
+    if (properties._cellLevel.fileFormat() == "h5" && options) {
         mut::Morphology mutable_morph(*this);
-        if (options) {
-            mutable_morph.applyModifiers(options);
-        }
+        mutable_morph.applyModifiers(options);
         properties_ = std::make_shared<Property::Properties>(mutable_morph.buildReadOnly());
         buildChildren(properties_);
     }


mgeplf:
Awesome!  Nice find.

We prob. need to have a check that the morph on disk was correct (which we are prob. not doing right now - but we need to have the proper order for h5 files)

matz-e:
> We prob. need to have a check that the morph on disk was correct (which we are prob. not doing right now - but we need to have the proper order for h5 files)

It would probably be good to traverse and check the order, and only rebuild when needed, too.

mgeplf:
> It would probably be good to traverse and check the order, and only rebuild when needed, too.

Yeah; I'd like to have time to clean this up at some point.

FYI; my (simple) benchmark gives a nice speedup to this too:
```
              old       new
loadH5      279926 ns   99673 ns
```

src/morphology.cpp
```diff
@@ -28,11 +28,9 @@ Morphology::Morphology(const Property::Properties& properties, unsigned int opti
 
     // For SWC and ASC, sanitization and modifier application are already taken care of by
     // their respective loaders
-    if (properties._cellLevel.fileFormat() == "h5") {
+    if (properties._cellLevel.fileFormat() == "h5" && options) {
```
mgeplf, line 31: very nice catch

mgeplf (approved):
I have done a conversion of asc -> h5 using:
```
morph-tool convert folder --extension h5 --recenter --sanitize asc h5_morphio_-options-optional
```

and the results are the same

I also tried w/o the `--sanitze --recenter`, and they are also the same

mgeplf:
I also tested with the NeuroM tests, and they all pass.

PR #393, opened 2022-03-18
Vector utils

mgeplf:


mgeplf:
@eleftherioszisis, @matz-e I'd appreciate some feedback if I can remove this (afaict) dead code.  I don't think it's particularly good to overload operators on alias types, b/c it causes other issues.  If anything we should create a proper Point, and use that; however, since none (except `subtract`) of the ops seem to be used in the code base, I think it's ok to keep the type alias, and just remove the operators.

Thoughts?

If you're good w/ this, I'd like to do some re-arranging after.

matz-e (commented):
Good from my side, does not look like MorphoKit is impacted.

eleftherioszisis:
I am in favor of removing non-used code and of reducing the ways things can go wrong.

include/morphio/vector_types.h
```diff
@@ -21,7 +19,7 @@ constexpr floatType PI = M_PI;
 /** Type of float to use. Can be double or float depending on MORPHIO_USE_DOUBLE */
 using floatType = float;
 /** A really small value that is used to measure how close are two values */
-constexpr floatType epsilon = 1e-6f;
+constexpr floatType epsilon = 1e-6F;
```
eleftherioszisis, line 22: Should it be upper case, given it's a constant?

eleftherioszisis (approved):
LGTM!

include/morphio/vector_types.h
mgeplf: good point; I'd like to remove it altogether - not a fan of having these sorts of things as constants 'everywhere'.  I'll think about it for another PR.

PR #398, opened 2022-04-05
Bump HighFive to v2.4.0

alkino:


alkino:
hdf5 1.8 is no more supported. Do you think you can upgrade it?

mgeplf (approved):

PR #403, opened 2022-05-11
Improve ascii morphology reading performance

NadirRoGue:
After running some quick benchmarks between the previous reading code (which reads char by char using istreambuf_iterator) and this method, which reads the whole file in blocks (maybe in a single read?), We have obtained the following reading measures:

Old method average time (milliseconds): 1299.375
New method average time (milliseconds): 6.71

The benchmark code is this:
```
#include <morphio/morphology.h>
#include <morphio/section.h>
#include <morphio/soma.h>

#include <chrono>
#include <filesystem>
#include <fstream>
#include <iostream>

int main(int, char**) {
    const auto path = std::filesystem::path(/* path to BBPTestData ascii morphology folder */);

    const auto files = std::filesystem::directory_iterator(path);

    auto start = std::chrono::high_resolution_clock::now();

    for (const auto& file : files) {
        const auto filePath = file.path().string();

        /*
        std::ifstream ifs(filePath, std::ios::ate);
        const auto size = ifs.tellg();
        ifs.seekg(0);
        std::string input;
        input.resize(static_cast<size_t>(size));
        ifs.read(&input[0], size);
        ifs.close();
        */

        std::ifstream ifs(filePath);
        std::string input((std::istreambuf_iterator<char>(ifs)),
                          (std::istreambuf_iterator<char>()));

        }
    auto end = std::chrono::high_resolution_clock::now();
    auto diff = end - start;
    auto millis = std::chrono::duration_cast<std::chrono::milliseconds>(diff);
    std::cout << "Time: " << millis.count() << " ms" << std::endl;
    return 0;
}
```


mgeplf:
Cool!  Let me investigate the `pip install --use-feature=in-tree-build` error separately.

NadirRoGue:
Just opened a new PR about that. They updated pip yesterday to 22.1, which removes that option and makes it default behaviour

NadirRoGue:
I dont fully get why the unittests are failing on Windows

mgeplf:
> I dont fully get why the unittests are failing on Windows

I'll have to look into it; adding it to the pile

NadirRoGue:
Apparently windows does not like this way of reading files?

mgeplf:
Reading a 2MB ASC, morphology, I get these numbers:
```
old;
BM_load_morphology/asc             26.9 ms

new:
BM_load_morphology/asc             20.9 ms
```
Using:
```
std::ifstream ifs(uri_);
std::ostringstream oss;
oss << ifs.rdbuf();
std::string input{oss.str()};
```
Looks safer to me, but isn't quite as fast: 
```
BM_load_morphology/asc             22.3 ms         22.3 ms           31
```

As @eleftherioszisis; points out `tellg` has some fun behavior: https://stackoverflow.com/a/22986486

NadirRoGue:
Thats a pity. I tried a 9 MB file, its 30x faster to read.
Would it be possible to modify MorphIO Morphology API, adding a constructor that can accept the data?
Brayns is a UNIX-only software, so we are safe to use tellg(), and we could greatly improve the performance on our side.

matz-e:
> Would it be possible to modify MorphIO Morphology API, adding a constructor that can accept the data?

Seconded, that would allow to store morphologies in a format better for GPFS :)

HDF5 already does that by allowing to pass a group from an open file rather than a path to a file.

mgeplf:
> Thats a pity. I tried a 9 MB file, its 30x faster to read.

Is that with the  `oss << ifs.rdbuf()` method?  It seems to be cross platform, and nearly as fast.

> Would it be possible to modify MorphIO Morphology API, adding a constructor that can accept the data?

Yeah, that's certainly something that could be done.

src/readers/morphologyASC.cpp
```diff
@@ -72,9 +72,19 @@ class NeurolucidaParser
...
+#ifdef _WIN32
         std::ifstream ifs(uri_);
         std::string input((std::istreambuf_iterator<char>(ifs)),
                           (std::istreambuf_iterator<char>()));
+#else
+        std::ifstream ifs(uri_, std::ios::ate);
+        const auto size = ifs.tellg();
+        ifs.seekg(0);
+        std::string input;
+        input.resize(static_cast<size_t>(size));
```
mgeplf, line 84: you can shave some more time off w/ `input.reserve` b/c then it doesn't need to fill the array, FYI.

NadirRoGue: The problem with reserve is that it does not modify the string size

mgeplf:
@NadirRoGue > Would it be possible to modify MorphIO Morphology API, adding a constructor that can accept the data?

I have added that here:
https://github.com/BlueBrain/MorphIO/pull/407

I also added faster reading, based on what you discovered with this PR; thanks!

NadirRoGue:
@mgeplf thanks a lot! I'll close this PR then.

PR #430, opened 2022-11-25
Fix failing tests for macos-latest that uses python3.11

eleftherioszisis:
With `macos-latest`, i.e. `macos-11.7.1` python was upgraded from 3.10 to 3.11:
https://github.com/actions/runner-images/commit/1f84dd7159610d519a3bdc2beff2352b41a9355f#diff-e68d4f6e862f03aa8800c0d5de1c1072dd2d01553ca0986d818f012c932a25a4L34
which results in trying to build h5py from source for the test runs, , given there are no wheels yet,  and herefore fails.

Setting up python to 3.10 for the tests solves this issue.

### DIFF
--- a/ci/python_test.sh
+++ b/ci/python_test.sh
@@ -1,4 +1,3 @@
-#!/usr/bin/env bash
 
 set -euxo pipefail
 
@@ -36,6 +35,7 @@ deactivate
 DIST_DIR="$VENV/dist"
 mkdir -p "$DIST_DIR"
 create_venv
+$BIN/pip install --no-binary h5py h5py
 python3 setup.py sdist --dist-dir "$DIST_DIR"
 $BIN/pip install "$DIST_DIR"/MorphIO*.tar.gz
 $BIN/pip install -r tests/requirement_tests.txt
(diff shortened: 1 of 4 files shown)


ci/python_test.sh
```diff
@@ -14,28 +14,32 @@ function create_venv()
...
+popd
+
 deactivate
 
 # sdist install
 DIST_DIR="$VENV/dist"
 mkdir -p "$DIST_DIR"
 create_venv
+
 python3 setup.py sdist --dist-dir "$DIST_DIR"
```
matz-e, line 39: For future reference, I think this is becoming deprecated, see [this](https://github.com/pytest-dev/pytest/issues/9225), pointing to [this](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

matz-e (approved):
Thank you, looks all pretty sensible!

ci/python_test.sh
eleftherioszisis: Nice! I switched to using that.

matz-e (approved):

PR #431, opened 2022-11-25
Add more options for ASC and SWC loaders

### LINKED ISSUE #427: Add a flag to disable some checks
MorphIO performs many checks when loading a file, but sometimes these checks are more restrictive than the standard (e.g. soma representation is more restrictive than the standard, it does not accept custom tree types, it does not accept bifurcations at the first point, etc.). It would be nice to be able to choose which level of checks are performed. We could start with just 2 levels: the min level for MorphIO to be able to load the file and the max level which is the current behavior. Later we could add a few other intermediate levels if it is needed.

adrien-berchet:
Fixes #427


adrien-berchet:
I did not add options for all checks, just for the ones that might be needed for TMD.

Also, I did not change the HDF5 reader at all for two reasons:
1. its implementation is deeply designed according to the specifications given in https://github.com/BlueBrain/morphology-documentation/blob/main/source/h5v1.rst, so it would be hard and dangerous to remove some checks.
2. the morphologies written in this format mainly come from MorphIO  which do not allow to write morphologies that do not follow its specifications, so I don't think we will ever find a HDF5 file containing a morphology that can not be loaded with the current loader.

adrien-berchet:
@eleftherioszisis and @mgeplf WDYT?

include/morphio/enums.h
```diff
@@ -11,11 +11,14 @@ enum LogLevel { ERROR, WARNING, INFO, DEBUG };
...
-    NO_DUPLICATES = 0x04,        //!< Skip duplicating points
-    NRN_ORDER = 0x08             //!< Order of neurites will be the same as in NEURON simulator
+    NO_MODIFIER = 0b0000000000000000,          //!< Read morphology as is without any modification
+    TWO_POINTS_SECTIONS = 0b0000000000000001,  //!< Keep only the first and last points of sections
+    SOMA_SPHERE = 0b0000000000000010,          //!< Interpret morphology soma as a sphere
+    NO_DUPLICATES = 0b0000000000000100,        //!< Skip duplicating points
+    NRN_ORDER = 0b0000000000001000,  //!< Order of neurites will be the same as in NEURON simulator
+    ALLOW_ROOT_BIFURCATIONS = 0b0000000000010000,  //!< Bifurcations at first point are allowed
+    ALLOW_SOMA_BIFURCATIONS = 0b0000000000100000,  //!< Bifurcations in soma are allowed
+    ALLOW_MULTIPLE_SOMATA = 0b0000000001000000     //!< Multiple somata are allowed
```
matz-e, line 21:
```suggestion
    NO_MODIFIER = 0,          //!< Read morphology as is without any modification
    TWO_POINTS_SECTIONS = 1 << 0,  //!< Keep only the first and last points of sections
    SOMA_SPHERE = 1 << 1,          //!< Interpret morphology soma as a sphere
    NO_DUPLICATES = 1 << 2,        //!< Skip duplicating points
    NRN_ORDER = 1 << 3,  //!< Order of neurites will be the same as in NEURON simulator
    ALLOW_ROOT_BIFURCATIONS = 1 << 4,  //!< Bifurcations at first point are allowed
    ALLOW_SOMA_BIFURCATIONS = 1 << 5,  //!< Bifurcations in soma are allowed
    ALLOW_MULTIPLE_SOMATA = 1 << 6     //!< Multiple somata are allowed
```
for readability?

src/readers/morphologyASC.cpp
```diff
@@ -139,9 +140,24 @@ class NeurolucidaParser
...
-            nb_.soma()->properties() = properties;
+            if (!nb_.soma()->points().empty()) {
+                if (options & ALLOW_SOMA_BIFURCATIONS) {
+                    printError(Warning::SOMA_BIFURCATION, err_.WARNING_SOMA_BIFURCATION());
+                } else if (options & ALLOW_MULTIPLE_SOMATA) {
+                    printError(Warning::MULTIPLE_SOMATA, err_.WARNING_MULTIPLE_SOMATA());
+                } else {
+                    throw SomaError(err_.ERROR_SOMA_ALREADY_DEFINED(lex_.line_num()));
+                }
+                nb_.soma()->properties()._points.insert(nb_.soma()->properties()._points.end(),
```
matz-e, line 151: There is no further distinction between bifurcation and multiple soma? They just get merged into one?

tests/CMakeLists.txt
```diff
@@ -48,10 +49,10 @@ target_link_libraries(unittests
...
 )
 
-add_test(NAME unittests
-         COMMAND unittests
-         WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
-         )
+# add_test(NAME unittests
+#          COMMAND unittests
+#          WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+#          )
```
matz-e, line 55: Why is this commented out?

adrien-berchet: Because I don't understand the purpose of this :)
As far as I could see, it just runs the tests again, which was a bit boring when I was testing things. But I can revert ofc, or just remove it if it's actually useless.

include/morphio/enums.h
adrien-berchet: Better indeed, thanks

src/readers/morphologyASC.cpp
adrien-berchet: For the ASC file format they appeared to be processed the same way. But for SWC it's different.

tests/CMakeLists.txt
adrien-berchet: For now I remove it, I can revert if you prefer.

src/readers/morphologyASC.cpp
matz-e: Ah, thanks for elaborating!

tests/CMakeLists.txt
matz-e: I think I would be more comfortable with that `add_test` still active. That `if` clause below may not always be triggered, and for me at least, I'd rather have the tests run twice than not at all.

adrien-berchet: Just to give more details, if you compare the CI logs before and after this change, you can see there is only one difference: before the change there was 1 more test at the end which was run with the following command: `Test command: /home/runner/work/MorphIO/MorphIO/bin/unittests`, i.e. without a test name given (others are run with commands like `Test command: /home/runner/work/MorphIO/MorphIO/bin/unittests "<test_name>"`). So it just runs all the tests at once after they have been run one by one.

But indeed if one sets `EXTERNAL_CATCH2` to `ON`, this might be needed. But in that case it would make more sense to move it in an `else` statement of the `if` you mentioned, isn't it?

matz-e: Yeah, in an `else` would be most appropriate

adrien-berchet: Ok, done

adrien-berchet:
@matz-e @eleftherioszisis @mgeplf 
Could you tell me if this PR looks relevant to you (we can speak about implementation details later)? If it does I will add some docs about these options. Thanks

eleftherioszisis:
> Could you tell me if this PR looks relevant to you (we can speak about implementation details later)? If it does I will add some docs about these options. Thanks
I am in favor of adding these options if they aid with processing morphologies with artifacts.

However, my main concern is about the repercussions of using these options and then converting between formats. I believe we need way more tests to ensure that the converters will not break when using these options.

adrien-berchet:
> However, my main concern is about the repercussions of using these options and then converting between formats. I believe we need way more tests to ensure that the converters will not break when using these options.

I think they will because the HDF5 format is very strict and I did not change much in this one because I think the HDF5 files are mainly created using MorphIO, even though they could be created in another way (and also because the HDF5 part of MorphIO is harder to change :innocent: ).
But I don't think it's really an issue if these flawed morphologies can not be written before they are fixed. But I am going to add some tests to see which cases are possible and which are not.

matz-e:
> I am in favor of adding these options if they aid in processing morphologies with artifacts.

> However, my main concern is about the repercussions of using these options and then converting between formats. I believe we need way more tests to ensure that the converters will not break when using these options.

Same. Also, this should not create any illusion that all our software can run with non-fixed morphologies. We have a pretty rigid idea of how a morphology should look like (exactly 1 soma, e.g.)

adrien-berchet:
> Same. Also, this should not create any illusion that all our software can run with non-fixed morphologies. We have a pretty rigid idea of how a morphology should look like (exactly 1 soma, e.g.)

That's why I added warnings in all irregular cases, but you think it's not enough?

To be more clear I improved the conversion test to check that morphologies that can be written can then be read without any specific option. This means that the irregular morphologies can be loaded with the new options BUT when they can be written (mainly for soma issues) they are automatically fixed (not in a smart way but just to make it work). This only happens for weird somata (e.g. multiple somata or somata with bifuractions). So the user gets a warning but if he is able to write it, whether he manually fixes the morphology or he lets MorphIO to fix it automatically if it is possible, then MorphIO will be able to load it with the default loader afterwards.

matz-e:
> but you think it's not enough?

I'm more concerned. Let's hope it's enough. People being aware helps a lot already.

adrien-berchet:
> > but you think it's not enough?
> 
> I'm more concerned. Let's hope it's enough. People being aware helps a lot already.

Ok :)

adrien-berchet:
Anything about this PR? @matz-e @mgeplf @eleftherioszisis

adrien-berchet:
@mgeplf @eleftherioszisis @matz-e
up :)
I rebased and it covers everything I need for now so for me it's ready.

mgeplf:
Thanks, I'm still far behind on fixing things :(  I will add it to the list of things to do....

PR #436, opened 2022-12-14
Replace assert by exception

### LINKED ISSUE #435: Assert used where an exception should be raised
https://github.com/BlueBrain/MorphIO/blob/aec82bfac2744bbae26be5f4ff49717e05c65eb3/morphio/__init__.py#L9

It's improper to use the `assert` keyword to raise an error when a library is mismatched. The `assert` keyword is to be used to assert internal code correctness. They may be optimized away with `-O` and `-OO` optimizer flags. It would be unfortunate if exactly those users looking to run under highly optimized conditions would run into a sneaky library mismatch bug :)

Helveg:
closes #435

### DIFF
--- a/morphio/__init__.py
+++ b/morphio/__init__.py
@@ -6,7 +6,8 @@ if platform.system() == "Windows":
     # there cannot be a mismatch between version, when using windows,
     # and MorphIO doesn't need to build hdf5
     import h5py as _h5py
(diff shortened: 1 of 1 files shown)


mgeplf (approved):

mgeplf:
Thanks for the fix, correctness is appreciated.

PR #439, opened 2023-02-09
Improve doc

### LINKED ISSUE #358: Add intersphinx feature
The Intersphinx extension allows to reference an object from another code in a documentation generated with Sphinx.
The doc of MorphIO is generated using Doxygen so it can not be cross linked with intersphinx by default. Maybe using this method would be nice to overcome this:
https://docs.typo3.org/typo3cms/extensions/sphinx/2.5.1/AdvancedUsersManual/AdvancedCrossLinks/Index.html#doxygen-documentation
So we could reference the morphio types in other documentations.

adrien-berchet:
Fixes #358 

The Python API part of the doc was not complete and had several formatting issues. This PR should fix these:

* Remove some warnings during doc generation
* Fix syntax highlighting in doc
* Improve doc for Python API
* Fix an assertion in a test that was always `True`


eleftherioszisis:
You have changes that are not reflecting this PR's description. Why did you remove the binding modules separated by context to join them in a single large and hard-to-read file?

adrien-berchet:
Ah yeah, this is because some methods of mutable classes need to be declared after immutable classes and some methods of immutable classes need to be declared after mutable classes. So all the classes must be defined before the methods. So if we want to keep things separated we need:

* a function that defines and returns mutable classes.
* a function that defines and returns immutable classes.
* a function that takes (almost) all classes as arguments to define the methods of mutable classes.
* a function that takes (almost) all classes as arguments to define the methods of immutable classes.

This was quite cumbersome so I gathered everything at the same place (and in the right order since the order matters). But we could try to split it again.

adrien-berchet:
Ah and just to explain a bit more: these methods need other classes to be defined before so that Sphinx is able to properly link the argument types or returned types of these methods.

adrien-berchet:
Well, I don't think it's less readable but I am going to try to split it again into several functions.

eleftherioszisis:
Before doing that, maybe @mgeplf could chip in.

adrien-berchet:
I split again the functions so it's more readable. It was not so easy because the `pybind11::class_` can not be constructed empty, so I had to store these objects into a struct. But in the end should be as readable as before.

tests/test_0_API.py
```diff
@@ -45,5 +45,18 @@ def methods(cls):
     assert (methods(morphio.EndoplasmicReticulum) ==
                  methods(morphio.mut.EndoplasmicReticulum))
 
-    assert (methods(morphio.Mitochondria),
-                 methods(morphio.mut.Mitochondria))
```
eleftherioszisis, line 49: These two sides are not equivalent, are they?

adrien-berchet: Not at all, that's why I changed the test. Previously the test was not properly written so it was spuriously passing.

eleftherioszisis: Spuriously? Apart from `append_root_section` the rest of the methods should be in both, no?
Also why there are the methods listed twice?

adrien-berchet: Look at the new test, all the methods listed in `only_in_mut_mitochondria` are different. I don't know why they are so different for this object though.

eleftherioszisis:
The documentation build from readthedocs is successful: https://readthedocs.org/projects/morphio/builds/19444990/

eleftherioszisis (approved):
lgtm. Thanks a lot for fixing and cleaning up the documentation!

PR #440, opened 2023-03-20
Make neurolucida state machine thread-safe

NadirRoGue:
The lexertl state machine used to parse the ascii morphologies isn't thread safe, as minimalisation of the machine can happen in parallel.
I splitted the build from the singleton to make the initialization thread-safe, while keeping the lazy initialization.


mgeplf (approved):

mgeplf:
makes sense, thanks for the change.

PR #442, opened 2023-04-13
Set locale to ensure dots are recognized as decimal points in SWC files

adrien-berchet:
In some cases, the function `sscanf` does not recognize dots as decimal points because of the `locale` settings of the user (french in my case). This PR ensures that dots are always considered as decimal points in SWC files, regardless of user locale settings.


mgeplf:
I'll look at this; I've wanted to remove the `sscanf` for a while; but I'm not super happy about how iostreams handle locales, either.  Plus, I think it was slower.

adrien-berchet:
Ah ok. As you prefer, this is just a quick fix but we could think about something better.

adrien-berchet:
Is #460 supposed to replace this PR @mgeplf ?

mgeplf:
Yeah, that's the plan.

adrien-berchet:
> Yeah, that's the plan.

Ok!

Issue #443, opened 2023-04-24
Add support for a storage container for many morphologies in one file


matz-e:
A while ago we added support for reading morphologies from groups in #163. This has led to experimental work by @1uc to see how well HDF5 files work to store many morphologies in a single file (presentation incoming). I think it would be good to consolidate the work done in our internal MorphoKit in MorphIO by

* Specifying how many morphologies should be stored in a single HDF5 file
  * the most simplistic approach by copying the contents of morphology files into sub-groups of a larger file. This layout is already facilitated by the group access in #163 and would be quick to implement.
  * concatenating all datasets of the morphologies into larger datasets and storing the offset, count into said datasets per morphology. This layout would allow to later perform similar optimizations as has been done for BlueBrain/libsonata by reading the required data for several morphologies in one go from the larger datasets.
* Adding a `MorphologyStorage` (or similar) class that takes either a directory or path to an HDF5 file and exposes a `.get` method to obtain morphologies stored
  * this can and should be expanded to provide optimized bulk access
  * at least a utility to create containers if not an interface to write should be included
  * this would be our new preferred way of reading morphologies for large circuits with a large number of unique morphologies
* Adding any other functionality in the reader required to support the format

mgeplf:
Everything above looks fine.  Please remember that this has to build under linux/mac and *windows*, since we produce wheels and such for all platforms.

> a `get` method to obtain morphologies stored

What would be the argument to the get?  A number? A string?

1uc:
The current solution in morpho-kit is essentially:
```
class MorphologyStorage {
public:
    virtual morphio::Morphology get(const std::string& morph_name) const = 0;
};

class DirectoryStorage {
public:
    DirectoryStorage(const std::string& dirname);
    morphio::Morphology get(const std::string& morph_name) const override;

public:
    const std::string& dirname;
};


class FileStorage {
public:
    FileStorage(const std::string& container_name);
    morphio::Morphology get(const std::string& morph_name) const override;

public:
    HighFive::File file;
};
```
The assumption is that the file name is `{dirname}/{morph_name}.{ext}` where there's some guessing an precedence order for picking the extension.

matz-e:
> Everything above looks fine. Please remember that this has to build under linux/mac and _windows_, since we produce wheels and such for all platforms.
> 
> > a `get` method to obtain morphologies stored
> 
> What would be the argument to the get? A number? A string?

[...]

PR #444, opened 2023-05-01
Abstraction for morphology collection.

1uc:
Currently, MorphIO requires the file, filename or HDF5 group which contains the morphology. This PR adds an interface for loading morphologies by name from a `Collection` object.

The first commits improve style, turn off problematic flags, add missing ctors and fix a bug in the HDF5 reader. Furthermore, they add gulrak's implementation of `std::filestystem`. The last commit adds the proposed feature and its tests.


1uc:
Please let me know if you want to split certain commits into their own PRs.

1uc:
This would implement part of #443.

include/morphio/collection.h
```diff
@@ -0,0 +1,82 @@
...
+#include <memory>
+#include <string>
+
+#include <morphio/morphology.h>
+#include <morphio/mut/morphology.h>
+
+namespace morphio {
+
+class CollectionImpl
+{
```
mgeplf, line 12: can we just forward decl this, and move the `load`s to `collection.cpp`?  I *try* (often unsuccessfully) to keep the headers as simple as possible.

include/morphio/collection.h
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Create a collection from the given path.
+     *
+     * If `collection_path` points to an HDF5 file, then that file
+     * must be a container. Otherwise the `collection_path` should
+     * point to the directory containing the morphology files.
+     *
+     * If the collection path is a directory, the extension of the morphology
+     * file must be guessed. The optional argument `extensions` specifies which
```
mgeplf, line 33: nice api; good idea

include/morphio/collection.h
```diff
@@ -0,0 +1,82 @@
...
+     * and in which order the morphologies are searched.
+     */
+    Collection(std::string collection_path,
+               std::vector<std::string> extensions = std::vector<std::string>{"h5", "swc", "asc"});
```
mgeplf, line 37: at bbp, I think we should put `swc` last; since it's generally not used.  Annoyingly, I'm not sure we're 100% standardized on lowercase `asc`, so adding `ASC` might be required :(  I will look around.

include/morphio/mut/morphology.h
```diff
@@ -41,7 +41,10 @@ class Morphology
     **/
     explicit Morphology(const std::string& uri, unsigned int options = NO_MODIFIER);
 
-    /// Build a mutable Morphology from a mutable morphology aaaaaa
```
mgeplf, line 44: ah yes, the classic `aaaaaa` :facepalm:

src/collection.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+        throw std::runtime_error("Morpholgy '" + morph_name + "' not found in: " + _dirname);
+    }
+
+  private:
+    std::string _dirname;
+    std::vector<std::string> _extensions;
+};
+
+class ContainerCollection: public morphio::detail::CollectionImpl<ContainerCollection>
```
mgeplf, line 58: could we add HDF5 to the name, just so it's clear what the backing store is?

src/readers/morphologyHDF5.cpp
```diff
@@ -318,13 +318,13 @@ void MorphologyHDF5::_read(const std::string& groupName,
...
-    if (!_group.exist(groupName)) {
+    if (groupName != "" && !_group.exist(groupName)) {
         throw(
             RawDataError("Reading morphology '" + _uri + "': Missing required group " + groupName));
     }
-    const auto group = _group.getGroup(groupName);
+    const auto group = groupName == "" ? _group : _group.getGroup(groupName);
 
-    if (!_group.exist(groupName)) {
+    if (!group.exist(datasetName)) {
```
mgeplf, line 327: is this the bug you reference in the PR description?

mgeplf (commented):
Looks good to me, I need to find some time to try it, but I think w/ a couple small changes, we'll be ready to merge.

include/morphio/collection.h
```diff
@@ -0,0 +1,82 @@
...
+     * Note that `morphio::Collection` uses RAII. Therefore, the usual
+     * scoping rules should suffice. However, there are cases where one
+     * wants to close the collection explicitly, causing all resources held
+     * by the collection to be release. In the case of containers it ensures
+     * that the file is closed.
+     *
+     * The object should not be used after calling `close`.
+     */
+    void close();
+
```
mgeplf, line 77: I wonder if a useful addition would be to have `list` or `contents` to get the available morphologies within the container?

src/readers/morphologyHDF5.cpp
1uc: Yes, when adding a test with a container that contained morphologies with perimeters, it would complain that there was to dataset called `"perimeters"`.

What makes me slightly wary is that we've used containers in TouchDetector. However, it's not impossible, that we've simply never used more complex morphologies with a perimeter.

The proposed change I believe is correct, whether the previous was a bug I'm not 100% sure. I strongly suspect it was, but as mentioned it would have been quite good at hiding.

CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 
 option(BUILD_BINDINGS "Build the python bindings" ON)
 option(MorphIO_CXX_WARNINGS "Compile C++ with warnings" ON)
+option(MorphIO_WERROR "If compiled with warnings, additionally activate '-Werror'" OFF)
```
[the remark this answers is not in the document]
1uc: That would lead to some double negation logic. We could just switch the default to `On`. Either is fine, as long as I can compile with warnings without them preventing progress by turning into errors.

include/morphio/collection.h
1uc: Good idea! I was thinking we'd get the morphology names from libsonata via a nodesfile; but that's not always the case. For directories it might be difficult to distinguish between files that contain morphologies and files that happen to have an extension that's used for morphologies. This gets slightly murkier since we often(?) nest morphologies in subdirectories of the morphology directory. The subdirectory then becomes part of the morphology name.

src/readers/morphologyHDF5.cpp
1uc: Sorry! I've not reread my description. Yes, the `datasetName` is relative to `groupName`. For fancy morphologies the `groupName` might be something like `"organelles/mitochondria"` and the `datasetName` would be `"points"`.

Hence, if we want to check that the dataset exists, we need to check the `group` (not `_group`) and check `datasetName`. It's highly likely it's a copy-pasting bug from the check on line 321.

1uc:
Thank you for the review. I'll change the unaddressed comments as suggested.

Currently, I'm working on integrating this into TouchDetector. We can try this "new" API (it's mostly copied from an internal library) there before merging it. This would check the C++ API and currently missing thread-safety (WIP).

After that we'll also need to prepare SpatialIndex. Which would check that the Python API is fine.

CMakeLists.txt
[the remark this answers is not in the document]
1uc: If `*_NO_WERROR` is your preferred name, I'm happy to change it. For now I only updated the default.

include/morphio/collection.h
1uc: Good point. Solved via explicit instantiation.

src/collection.cpp
1uc: done.

include/morphio/collection.h
1uc: @mgeplf Can this please be a separate task? I'd like to prepare the migration for two of our internal applications; but their CI will only run after this has been merged. Therefore, it would be helpful to limit the functionality of this PR.

So far I've not needed know what's inside a collection, I've always had a nodesfile tell me what I need to load.

mgeplf: yeah, that's no problem

We'll need a way to list what's there, at least of h5 files, so that we can quickly reconcile that with nodes files, if things aren't working - but we can do that in the next step.

mgeplf:
I'm basically happy with this.  Is there anyone else you'd like to review it?

1uc:
I've run it through TD and that works. I'm checking SI.

CMake/CompilerFlags.cmake
```diff
@@ -12,7 +12,10 @@ if (WIN32)
 elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   # Taken from https://github.com/BlueBrain/hpc-coding-conventions/blob/master/cpp/cmake/bob.cmake#L192-L255
   if(${PROJECT_NAME}_CXX_WARNINGS)
-    set(FLAGS "${FLAGS} -Werror -Weverything")
+    if(${PROJECT_NAME}_WERROR)
```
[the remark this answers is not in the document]
1uc: I think it's good to be able to see warnings while developing. Often they're just that warnings, some are true some are false. Yes, one should work towards eliminating them all; otherwise one misses the important one.

My issue with `-Werror` is that it makes appeasing the compiler priority number one. Nothing else is possible until there are no more warnings.

Concretely, GCC 12.2.1 will emit many warnings from within third-party libraries.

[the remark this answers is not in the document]
mgeplf: Ok, you've convinced me; we'll keep it.  Thanks.

mgeplf:
I've tried merging this several times, but w/ the github outage, I think it's best I wait until tomorrow.

PR #447, opened 2023-05-10
Implement native support for unified containers

1uc:
Unified HDF5 morphology containers internally concatenate the datasets across all morphologies.


src/readers/morphologyHDF5.cpp
```diff
@@ -78,9 +78,165 @@ namespace morphio {
...
+static void check_dimensions_match(const std::vector<size_t>& actualDimensions,
+                                   const std::vector<size_t>& expectedDimensions,
+                                   const std::string& datasetName) {
+    if (actualDimensions.size() != expectedDimensions.size()) {
+        throw RawDataError("dataset: " + datasetName + ": bad number of dimensions.");
+    }
+
+    for (size_t k = 0; k < actualDimensions.size(); ++k) {
+        size_t expected = expectedDimensions[k];
+        size_t actual = size_t(actualDimensions[k]);
```
mgeplf, line 104: is the `size_t` needed?
```suggestion
        size_t actual = actualDimensions[k];
```

src/readers/morphologyHDF5.h
```diff
@@ -12,6 +12,92 @@ namespace h5 {
 Property::Properties load(const std::string& uri);
 Property::Properties load(const HighFive::Group& group);
 
+namespace detail {
+/**
+ * Provides the API for reading datasets from HDF5 files.
+ *
+ * The CRTP requirements are that `Derived` has a method
```
mgeplf, line 19: do we need to CRTP vs virtual, and rely on devirtualization? Is this a hot dispatch?

src/readers/morphologyHDF5.h
```diff
@@ -35,9 +121,19 @@ class MorphologyHDF5
...
+    void _read(const std::string& group,
+               const std::string& _dataset,
+               const std::vector<size_t>& expectedDimensions,
+               T& data);
+
     HighFive::Group _group;
     Property::Properties _properties;
     std::string _uri;
+
+    int _file_version;
```
mgeplf, line 134: I take it we're making the assumption that all morphologies within a container have the same version; this makes sense, but perhaps we should write that down somewhere (I assume I've missed it :()

mgeplf (commented):
looks like the right direction to me

src/readers/morphologyHDF5.cpp
```diff
@@ -78,9 +78,165 @@ namespace morphio {
 namespace readers {
 namespace h5 {
 
+
+namespace detail {
+
+// Note, that `hsize_t` might be `uint64_t` which could with `uint64_t` or
```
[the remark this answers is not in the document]
1uc: I should learn to write :)

src/readers/morphologyHDF5.cpp
1uc: not anymore.

src/readers/morphologyHDF5.h
1uc: This is the version of the container, not of individual morphologies. "merged" is 3 and "unified" is 4. One can mix morphologies of different versions, e.g. v1.x, which is checked in the unit-tests, both `merged.h5` and `unified.h5` include `"simple.h5", "mitochondria.h5", "endoplasmatic-reticulum.h5"`. I'll need to add `simple-dendritric-spine.h5`.

src/readers/morphologyHDF5.h
1uc: I think we can't use virtual function calls, because of the `T& out`. This is usually `std::vector<T>& out` but the inner type differs. I don't think this goes away even if we were to switch to `T read(...) const`. Therefore, the only way I know to avoid a bit of code duplication is to use CRTP.

PR #448, opened 2023-05-11
Prefer exception over assert.

1uc:
This assert seems generally relevant and should be fast compared to the rest of the function which includes disk access.


mgeplf (approved):

mgeplf:
thanks

PR #450, opened 2023-05-15
Allow options in `Collection.load`.

1uc:
This PR adds the ability to pass modifiers when loading morphologies via the Collection API. Moreover, it updates the file `merged.h5` to include the test file required to check that NRN order is obeyed; and with a morphology with dentritic spines. Finally test of the container are added to include a H5 v1.3 morphology.


1uc:
There changes are required to add container support to neuron via morphio.

mgeplf (approved):

PR #451, opened 2023-05-17
Prevent `pytest` from picking up pybind11 tests.

1uc:
When simply running `pytest` after a performing `pip install -e .` I get the following errors:
```
binds/python/pybind11/tests/conftest.py:16: in <module>
    import pybind11_tests
E   ModuleNotFoundError: No module named 'pybind11_tests'
```
which means it's picking up the pybind11 tests and fails. This PR proposes to limit the folders pytest considers to `tests`.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+testpaths =
+    tests


1uc:
The PR shows two ways of solving the issue either via `pyproject.toml` or with a `pytest.ini` file.

eleftherioszisis (approved):
lgtm. pyproject.toml is the way to go for restricting pytest discovery.

mgeplf (approved):

PR #452, opened 2023-05-22
Don't ignore choice of `MORPHIO_ENABLE_COVERAGE`.

1uc:
This commit enables specifying `MORPHIO_ENABLE_COVERAGE` from the CLI. This is useful when building MorphIO in an IDE which may not have access to `lcov` and subsequently fail for `Debug` builds. I think it's sensible to want debug flags, but not be interested in the coverage. If so, `lcov` should be an optional dependency.


mgeplf:
tnx, btw, I like it

Issue #457, opened 2023-06-20
Create warning when an incompatible soma is trying to be written


mgeplf:
* SWC files should have soma that are either a point, or a set of stacked cylinders.
* H5 & ASC consider the soma to be a contour

Let's make this raise a warning in the code.  This should become a hard error in the future.

mgeplf:
Fixed by #458

PR #458, opened 2023-06-20
Have warnings for incompatible soma-types

mgeplf:
* SWC files should have soma that are either a point, or a set of stacked cylinders.
* H5 & ASC consider the soma to be a contour


mgeplf:
fixes #457

binds/python/generated/docstrings.h
```diff
@@ -23,6 +23,54 @@
 #endif
 
 
+static const char *__doc_morphio_Collection = R"doc()doc";
```
mgeplf, line 26: Skip reading this, it's autogenerated

eleftherioszisis (approved):
LGTM. I like how things are strict now wrt type/soma points.

PR #461, opened 2023-06-22
Increase coverage

mgeplf:


binds/python/generated/docstrings.h
```diff
@@ -23,6 +23,54 @@
 #endif
 
 
+static const char *__doc_morphio_Collection = R"doc()doc";
```
mgeplf, line 26: no need to review this, it's machine generated

src/properties.cpp
```diff
@@ -7,80 +7,10 @@
...
-    }
-
-    this->_points = other._points;
-    this->_diameters = other._diameters;
-    this->_perimeters = other._perimeters;
-    return *this;
-}
-
-template <typename T>
-bool compare(const std::vector<T>& vec1,
```
mgeplf, line 54: the `compare()` functions never seem to be used, so I removed them

eleftherioszisis (approved):

PR #466, opened 2023-07-03
remove excessive static_casts in SWC reader

mgeplf:
* don't allow negative IDs, except for the special parent == -1, for the first sample

### DIFF
--- a/src/readers/morphologySWC.cpp
+++ b/src/readers/morphologySWC.cpp
@@ -1,5 +1,6 @@
 #include "morphologySWC.h"
 
+#include <algorithm>
 #include <cstdint>        // uint32_t
 #include <memory>         // std::shared_ptr
 #include <sstream>        // std::stringstream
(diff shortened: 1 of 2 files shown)


src/readers/morphologySWC.cpp
```diff
@@ -14,30 +15,49 @@
 #include <morphio/properties.h>
 
 namespace {
+// It's not clear if -1 is the only way of identifying the root section.
+const int SWC_UNDEFINED_PARENT = -1;
+const unsigned int SWC_ROOT = 0xFFFFFFFD;
```
eleftherioszisis, line 20: Does this need to be a different "unknown" number? Could we also use here `0xFFFFFFFE` or is it required they are differentiated?

mgeplf: doesn't need to be different, but it helps w/ debugging, so I like to use different numbers.

src/readers/morphologySWC.cpp
```diff
@@ -14,30 +15,49 @@
...
                           format,
-                          &sample.id,
+                          &id,
                           &int_type,
                           &sample.point[0],
                           &sample.point[1],
                           &sample.point[2],
                           &radius,
-                          &sample.parentId) == 7;
+                          &parentId) == 7;
```
eleftherioszisis, line 47: Maybe turn 7 into a constant?

mgeplf: this will be going away once I get rid of `scanf`

eleftherioszisis (approved):

PR #476, opened 2023-10-16
Efficient swc build

mgeplf:
New SWC parser:
 * Parsing an SWC goes from 1.01ms to 0.700ms
 * Supports locales


src/readers/morphologySWC.cpp
```diff
@@ -53,397 +162,374 @@ std::vector<unsigned int> gatherLineNumbers(const std::vector<SWCSample>& sample
...
-       and update the parent ID of the new section
-    */
-    void _processSectionStart(mut::Morphology& morph, const SWCSample& sample) {
-        Property::PointLevel properties;
+        const SWCSample* sample = &samples_.at(id);
 
-        uint32_t id = 0;
+        // create duplicate point if needed
+        if (!is_root && sample->point != start_point /*|| sample->diameter != start_diameter */) {
+        //if (!(is_root && sample->type == SECTION_SOMA) && sample->point != start_point /*|| sample->diameter != start_diameter */) {
```
eleftherioszisis, line 470: Is this intentionally commented out?

src/readers/utils.cpp
```diff
@@ -0,0 +1,78 @@
...
+#endif
+
+StringToNumber::StringToNumber()
+#if defined(_WIN32) || defined(__WIN32__) || defined(WIN32)
+    : locale(_create_locale(LC_ALL, "C")) {
+}
+#else
+    : locale(newlocale(LC_NUMERIC_MASK, "POSIX", nullptr)) {
+}
+#endif
```
eleftherioszisis, line 31: Same here. It's not clear to me why these are configured this way.

eleftherioszisis (approved):

src/readers/utils.cpp
mgeplf: done

src/readers/morphologySWC.cpp
mgeplf: fixed, thanks.

PR #481, opened 2023-11-23
add cache of hdf5 and libaes source

mgeplf:


.github/workflows/publish-sdist-wheels.yml
```diff
@@ -25,69 +27,70 @@ jobs:
         with:
           submodules: 'true'
 
-      - uses: actions/setup-python@v4
-        name: Install Python
+      - name: Install Python
+        uses: actions/setup-python@v4
         with:
-          python-version: '3.7'
```
eleftherioszisis, line 31: Why not just jump to 3.10 so that we don't have to change it soon?

.github/workflows/publish-sdist-wheels.yml
```diff
@@ -25,69 +27,70 @@ jobs:
...
-              --enable-build-mode=production \
-              --with-szlib                   \
-              --enable-tools=no              \
-              --enable-tests=no
-            make -j 2
-            make install
-            ldconfig
-            popd
+
+            # CMake complains if the dependencies come from within the same tree
```
eleftherioszisis, line 56: lol

.github/workflows/publish-sdist-wheels.yml
mgeplf: good call

PR #483, opened 2023-12-06
break up writers.cpp into individual writers

mgeplf:
* unify duplicate functionality into writer_utils


eleftherioszisis (approved):
nice!

PR #493, opened 2024-04-19
follow libsonata's example, and publish wheels that statically use hdf5, to reduce h5py version conflicts

mgeplf:


eleftherioszisis (approved):
LGTM

PR #501, opened 2024-09-04
Update upload/download actions to `v4`.

1uc:


matz-e (approved):

.github/workflows/publish-sdist-wheels.yml
```diff
@@ -149,9 +149,9 @@ jobs:
           python -m cibuildwheel --output-dir dist
 
       - name: Store wheel as artifact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: dist
+          name: dist-wheels
```
mgeplf, line 154: I wonder if this should be made "more" unique using `${{github.sha}}` since I got this failure:
https://github.com/BlueBrain/MorphIO/actions/runs/10735932447/job/29774274616#step:12:27

1uc: Yes, you need to add `dist-wheels-${{ matrix.os }}` or something similar.

mgeplf: Thanks, I'll do that.

PR #505, opened 2024-10-08
update HDF5 url

mgeplf:
* now it's some random S3 bucket, because the https://support.hdfgroup.org/ftp/HDF5/ version doesn't seem to be curl-able


mgeplf:
I got the url from https://support.hdfgroup.org/downloads/hdf5/hdf5_1_14_3.html

1uc (approved):

PR #510, opened 2024-10-11
Section change

mgeplf:
* A section within MorphIO is defined as a series of segments between bifurcation/multifurcation points.  
  However, SWC files allow section change without a branch.
  Normal parsing of this will raise an exception, but can be allowed using the option `UNIFURCATED_SECTION_CHANGE`

### DIFF
--- a/include/morphio/enums.h
+++ b/include/morphio/enums.h
@@ -20,7 +20,7 @@ enum Option {
     SOMA_SPHERE = 0x02,          //!< Interpret morphology soma as a sphere
     NO_DUPLICATES = 0x04,        //!< Skip duplicating points
     NRN_ORDER = 0x08,            //!< Order of neurites will be the same as in NEURON simulator
-    UNIFURCATED_SECTION_CHANGE = 0x10  //!< Allow section type to change without bifurcation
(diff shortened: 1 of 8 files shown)


include/morphio/enums.h
```diff
@@ -19,7 +19,8 @@ enum Option {
     TWO_POINTS_SECTIONS = 0x01,  //!< Read sections only with 2 or more points
     SOMA_SPHERE = 0x02,          //!< Interpret morphology soma as a sphere
     NO_DUPLICATES = 0x04,        //!< Skip duplicating points
-    NRN_ORDER = 0x08             //!< Order of neurites will be the same as in NEURON simulator
+    NRN_ORDER = 0x08,            //!< Order of neurites will be the same as in NEURON simulator
+    UNIFURCATED_SECTION_CHANGE = 0x10  //!< Allow section type to change without bifurcation
```
eleftherioszisis, line 23: Would it be too big if instead we had `ALLOW_UNIFURCATED_SECTION_SCHANGE`?

mgeplf: sounds good to me, will do

eleftherioszisis (approved):
LGTM

PR #513, opened 2024-11-11
Prepare for HighFive v3.

1uc:


1uc:
@mgeplf what does one do about the coverage drop? It seems like the usual issue of reducing the total number of lines by removing covered lines, leads to a decrease in coverage (by numbers).

mgeplf:
> what does one do about the coverage drop? It seems like the usual issue of reducing the total number of lines by removing covered lines, leads to a decrease in coverage (by numbers).

it's an annoying effect of using the ratio, for sure.
I personally add a couple tests, since so much is not covered at the moment.
I don't expect you to do that, though - I will have a look.

PR #515, opened 2024-11-21
Fix a test regression with pybind11 2.11.2, 2.12.1, 2.13.6+

musicinmybrain:
“A new `self._pybind11_conduit_v1_()` method is automatically added to all `py::class_`-wrapped types, to enable type-safe interoperability between different independent Python/C++ bindings systems, including pybind11 versions with different `PYBIND11_INTERNALS_VERSION`s.”

https://github.com/pybind/pybind11/releases/tag/v2.13.6

Ensure that `tests/test_0_API.py::test_doc_exists` does not see this as an undocumented API function.

### DIFF
--- a/tests/test_0_API.py
+++ b/tests/test_0_API.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 import morphio
 
-
 def test_doc_exists():
     classes = [
         morphio.EndoplasmicReticulum,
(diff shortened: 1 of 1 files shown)


mgeplf (commented):
Thanks; we had discussed this in another PR, but forgot to break it out into a separate request.

I will update this PR to include what was decided there.

mgeplf (approved):

PR #517, opened 2025-01-15
Include <cstdint> for fixed-width integers in API headers

musicinmybrain:
At least in the API headers, include `<cstdint>` everywhere that fixed-width integer types `int32_t`, `uint32_t`, or `uint64_t` are used. Including headers everywhere they are used avoids relying on indirectly including them via the C++ standard library, which tends to break with new compiler versions. This change doesn’t implement that idea project-wide even for `<cstdint>`, let alone in general, but it is enough to fix failure to build with GCC 15 (as observed in Fedora).

### DIFF
--- a/include/morphio/mitochondria.h
+++ b/include/morphio/mitochondria.h
@@ -4,6 +4,7 @@
  */
 #pragma once
 
+#include <cstdint>  // uint32_t
 #include <memory>
 #include <vector>
 
--- a/include/morphio/morphology.h
+++ b/include/morphio/morphology.h
@@ -4,6 +4,8 @@
  */
 #pragma once
 
+#include <cstdint>  // uint32_t
+
 #include <highfive/H5Group.hpp>
 
 #include <morphio/properties.h>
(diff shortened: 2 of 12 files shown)


matz-e (approved):
Thanks! BTW, I _think_ that future development of this has moved into the @openbraininstitute organization?

WeinaJi:
> Thanks! BTW, I _think_ that future development of this has moved into the @openbraininstitute organization?

Yes, it is now https://github.com/openbraininstitute/morphio @eleftherioszisis

musicinmybrain:
> Thanks! BTW, I _think_ that future development of this has moved into the @openbraininstitute organization?

> Yes, it is now https://github.com/openbraininstitute/morphio @eleftherioszisis

Ah, thank you!

1. I replicated this PR at https://github.com/openbraininstitute/MorphIO/pull/1.
2. I updated the URL in Fedora’s [`morphio` package](https://src.fedoraproject.org/rpms/morphio) – at least in Rawhide / future Fedora 42.
3. I verified that https://release-monitoring.org/project/60593/ is tracking PyPI releases and therefore doesn’t need updating.

